PR #9, opened 2018-01-18
User Content API and other improvements.

m-mohr:
Improved User Content API; added OpenSearch response examples and switched to Atom as response format; minor alignments, improvements and fixes.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -3,7 +3,7 @@
   "info": {
     "title": "OpenEO API",
     "version": "0.0.1",
-    "description": "The OpenEO API specification for interoperable cloud-based processing of large Earth observation datasets. **This early draft version is incomplete and intended for working on a prototype and a proof of concept.** Things that are currently missing particularly include: \n * Authentication and authorization with OAuth 2.0, \n * how results of computations can be downloaded, \n * how data is streamed into UDFs and how the output of UDFs is returned, \n * how services are organized as microservices, \n * and  how OpenSearch is interfaced",
+    "description": "The OpenEO API specification for interoperable cloud-based processing of large Earth observation datasets. **This early draft version is incomplete and intended for working on a prototype and a proof of concept.** Things that are currently missing particularly include: \n * Authentication and authorization with OAuth 2.0, \n * how results of computations can be downloaded, \n * how data is streamed into UDFs and how the output of UDFs is returned, \n * how services are organized as microservices, \n * how payments can be handled, \n * and how OpenSearch is interfaced",
     "contact": {
       "name": "Marius Appel",
       "url": "https://github.com/open-eo/openeo-api-poc",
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -523,7 +523,7 @@
               "text/plain; charset=utf-8": "b34ba2bdf9ac9ee1"
             }
           },
-          "404": {
+          "403": {
```
m-mohr: I think 403 is more meaningful here in case of a failed authentication.

swagger.json
```diff
@@ -622,14 +622,12 @@
     "/data/opensearch": {
       "get": {
         "summary": "OpenSearch endpoint to receive standardized data search results.",
-        "description": "This service offers more complex search functionality and returns results in an OpenSearch compliant RSS XML format.",
+        "description": "This service offers more complex search functionality and returns results in an OpenSearch compliant Atom XML format.",
```
m-mohr: Atom seems more flexible, e.g. you can add information about source/rights, and is basically the successor of RSS. In case we want to allow XML, Atom AND RSS, we should remove the specific information here.

swagger.json
```diff
@@ -622,14 +622,12 @@
...
         "tags": [
           "EO Data Discovery"
         ],
         "produces": [
-          "application/rss+xml",
-          "application/atom+xml",
-          "application/xml"
+          "application/atom+xml"
```
m-mohr: See above.

swagger.json
```diff
@@ -1820,7 +1824,7 @@
           {
             "name": "lang",
             "in": "path",
-            "description": "Language identifier such as `'R'`",
```
m-mohr: The code environment is already highlighting the data with another font family. It also encourages users to think the quotes are part of the response/code/... and need to be included. Changed that throughout the whole file.

swagger.json
```diff
@@ -1589,13 +1584,22 @@
...
           },
           "404": {
-            "description": "File with specified path is not available"
+            "description": "File with specified path is not available."
+          },
+          "415": {
+            "description": "File type that is being uploaded is not allowed."
+          },
+          "423": {
+            "description": "The file that is being accessed is locked."
```
m-mohr: There are probably more potential "errors" we need to take care of.

swagger.json
```diff
@@ -1450,7 +1448,7 @@
         ],
         "responses": {
           "200": {
-            "description": "Flattened file tree with path relative to the user's root directory and some basic properties such as file size.",
+            "description": "Flattened file tree with path relative to the user's root directory and some basic properties such as the file size and the timestamp of the last modification. All properties except the name are optional.",
```
m-mohr: We should clearly state which attributes can be supported and how they are formatted (e.g. timestamps as ISO8601?).

swagger.json
```diff
@@ -219,22 +219,7 @@
       ]
     },
     "udf_description": {
-      "description": "Defines and describes a UDF using the same schema as the description of processes offered by the back-end.",
```
m-mohr: When using $ref all other arguments are ignored. If we want the custom description and example we need to copy over the other content from the referenced definition.

Issue #12, opened 2018-01-23
Add CORS headers for the JS client


m-mohr:
The openEO JS Client can't access any backend due to the CORS problem in client-side JavaScript.
Therefore every backend should send the `Access-Control-Allow-Origin: *` header.
This should be added to the API specification. See https://enable-cors.org/server.html for more information.

Maybe we also need to set the `Access-Control-Allow-Headers` header, allowing `Authorization, Accept, Content-Type` and whatever we need according to the API specification.  The background is that with custom headers, like the `Authorization` header we use for login a preflight (using the `OPTIONS` header) is made by the browsers, which might abort the following request (e.g. `GET`). This leaded to problems during implementation when I tried to authenticate as the `GET` request was never sent by the browser.

According to https://zinoui.com/blog/cross-domain-ajax-request we might also need to set `Access-Control-Allow-Credentials` and `Access-Control-Allow-Methods`.

Note: A wildcard `*` cannot be used in the `Access-Control-Allow-Origin` header when the credentials flag is true.

m-mohr:
Tackled it with #16, but we still need to think about authorization and the corresponding headers, e.g. `Access-Control-Allow-Credentials`.

m-mohr:
Authorization is added in #16, too. This should now fully implement everything related to CORS issues.

m-mohr:
Implemented with PR #16.

PR #16, opened 2018-02-05
CORS and authorization

m-mohr:
See issue #12.

**Merge #14 first!**

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -83,6 +83,9 @@
     },
     "unavailable": {
       "description": "The service is currently unavailable."
+    },
+    "method_not_allowed": {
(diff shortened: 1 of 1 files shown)


appelmar:
Would it make sense to add schema definitions for "Access-Control-Allow-Origin" and "Content-Type" response headers in order to minimize redundancy? I also found that Open API 3.0.0 recommends: "_If a response header is defined with the name "Content-Type", it SHALL be ignored._"

m-mohr:
I'm still working on it due to the missing authorization headers. It get's even more complicated, but I try to reduce redundancy with the next commit.

PR #21, opened 2018-02-06
Mostly minor clarifications and improvements, added PUT/PATCH methods to potentially change jobs, services, ...

m-mohr:
Please have a look at the two PATCH methods introduced. I'll comment on them in the diff to make them stand out in the changes. Their specifications are mostly placeholders for now, but it seemed a good idea to have them, also in the context of "restarting" lazy jobs.

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -1582,6 +1582,54 @@
           }
         }
       },
+      "get": {
+        "summary": "Lists user-defined processes.",
+        "description": "This service lists all processes of the user that are stored at the back-end.",
+        "tags": [
+          "User Content"
+        ],
+        "security": [
+          {},
+          {
+            "Bearer": []
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "JSON array with process_ids",
+            "examples": {
+              "application/json": [
+                "57d7e8ff91a1134",
+                "12bd7584efa1236"
+              ]
+            },
+            "headers": {
+              "Access-Control-Allow-Origin": {
+                "type": "string",
+                "description": "Allowed origin of the request, usually `*`",
+                "default": "*"
+              },
+              "Access-Control-Allow-Credentials": {
+                "type": "string",
+                "description": "If authorization is implemented by the back-end it needs to be set to `true`.",
+                "default": "true"
+              }
+            }
+          },
+          "401": {
+            "$ref": "#/responses/auth_required"
+          },
+          "403": {
+            "$ref": "#/responses/access_denied"
+          },
+          "501": {
+            "$ref": "#/responses/not_implemented"
+          },
+          "503": {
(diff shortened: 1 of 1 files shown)


swagger.json
```diff
@@ -2009,6 +2046,92 @@
           }
         }
       },
+      "patch": {
```
m-mohr: Changing a web service. The type and job_id should not be changed (therefore removed), but the configuration parameters (args) might be needed to changed without creating a new service id. Otherwise every change would trigger a new URL for (shared) services.

swagger.json
```diff
@@ -2206,8 +2337,80 @@
...
```
m-mohr: Changing a job. The process graph might be needed to be changed without creating a new job id. This sounds especially useful as we removed the DELETE endpoint for jobs. Otherwise every change would trigger a new URL for (shared) jobs. We still need to think about what happens with running jobs when patching/changing. etc.

swagger.json
```diff
@@ -2527,7 +2743,7 @@
       },
       "get": {
         "summary": "Cancels any back-end computations of a job",
-        "description": "For batch jobs this request cancels all related computations at the back-end. Lazy jobs will not respond to any further on demand requests, e.g. download calls or service requests. This job won't generate additional costs for processing. Results of batch jobs might be discarded and deleted by the back-end.",
+        "description": "For batch jobs this request cancels all related computations at the back-end. Lazy jobs will not respond to any further on demand requests, e.g. download calls or service requests. This job won't generate additional costs for processing unless started again using `GET /jobs/{job_id}/queue` (for batch jobs) or updated using `PATCH /jobs`. Results of batch jobs might be discarded and deleted by the back-end.",
```
m-mohr: We didn't have a solution to "restart" a lazy job. Does this clarification sound useful?
So restarting a lazy job could be done with a (maybe empty) patch.

swagger.json
```diff
@@ -1562,9 +1582,57 @@
           }
         }
       },
+      "get": {
```
m-mohr: We completely missed the backend to get a list of all user-defined process graphs.

swagger.json
```diff
@@ -1684,9 +1767,64 @@
           }
         }
       },
+      "put": {
```
m-mohr: There should be a possibility to replace the process graph and maintain the existing ID.

swagger.json
```diff
@@ -2206,8 +2444,80 @@
...
```
m-mohr: Replacing a job. The process graph might be needed to be changed without creating a new job id. This sounds especially useful as we removed the DELETE endpoint for jobs. Otherwise every change would trigger a new URL for (shared) jobs. We still need to think about what happens with running jobs when patching/changing. etc.

PR #22, opened 2018-02-06
Adopted the changes introduced in API spec. v0.0.2 and the glossary into the docs.

m-mohr:


m-mohr:
Merged the glossary into the docs.

docs/docs/glossary.md
```diff
@@ -0,0 +1,70 @@
...
+
+The terms _process_, _process graph_ and _job_ have different meanings in the OpenEO API specification.
+
+A **process** is simply the description of an operation as provided by the back end, similar to a function definition in programming languages. 
+
+A **process graph** includes specific process calls, i.e. references to one or more processes including specific values for input arguments similar to a function call in programming. However, process graphs can chain multiple processes. In particular, arguments of processes in general can be again (recursive) process graphs, input datasets, or simple scalar or array values.
+
+A **job** brings one process graph to the back-end and organizes its execution, which may or may not induce costs. Jobs furthermore allow to run process graphs from different _data views_ (see section on [data views](views.md)). Views define at which resolution and extent we look at the data during processing and hence allow to try out process graphs on small subsets, or work interactively within web map applications. For more information about jobs and their evaluation types, see the section on [jobs](jobs.md).
+
+In this context OpenEO will:
```
christophfriedrich: This chunk of text ("In this context OpenEO will" + the 4 bullet points after it) is out of context. I even think that such a feature description shouldn't be part of a glossary.

docs/docs/glossary.md
```diff
@@ -0,0 +1,70 @@
...
+
+Examples:
+
+- a time series aggregation may return a regression slope and intercept for every pixel time series, for a single band (group by: full time extent)
+- a time series may be aggregated to monthly values by computing the mean for all values in a month (group by: months)
+- _spatial_ aggregation involves computing e.g. _mean_ pixel values on a 100 x 100 m grid, from 10 m x 10 m pixels, where each original pixel is assigned uniquely to a larger pixel (group by: 100 m x 100 m grid cells)
+
+Note that for the first example, the aggregation function not only requires time series values, but also their time stamps.
+
+***Resampling*** is a broader term where we have data at one resolution, and need values at another (also called _scaling_). In case we have values at a 100 m x 100 m grid and need values at a 10 m x 10 m grid, the original values will be reused many times, and may be be simply assigned to the nearest high resolution grid cells ("nearest neighbor"), or may be interpolated somehow (e.g. by bilinear interpolation). Resampling from finer to coarser grid by nearest neighbor may again be a special case of aggregation.
```
christophfriedrich: Resampling from finer to coarser grid by techniques other than "nearest neighbor" may be a special case of aggregation too, may it not?

docs/docs/glossary.md
```diff
@@ -0,0 +1,70 @@
...
+- **client**: software environment (software) that end-users directly interact with, e.g. R (rstudio), python (jupyter notebook), and javascript (web browser); R and python are two major data science platforms; javascript is a major language for web development
+- **(cloud) back-end**: server; computer infrastructure (one or more physical computers or virtual machines) used for storing EO data and processing it
+- **big Earth observation cloud back-end** server infrastructure where industry and researchers analyse large amounts of EO data
+- **simple** many end-users now use python or R to analyse data and javascript to develop web applications; analysing large amounts of EO imagery should be equally simple, and seamlessly integrate with existing workflows
+- **unified** current EO cloud back-ends all have [a different API](http://r-spatial.org/2016/11/29/openeo.html), making EO data analysis hard to validate,difficult to reproduce, and back-ends difficult to compare in terms of capability and costs, or to combine in a joint analysis across back-ends. A unified API can resolve many of these problems.
```
christophfriedrich: "unified" isn't really explained here, instead a rundown of "what it's not" is given.
"simple" isn't really *defined* either.
IMO the two definitions should be replaced with a proper one (I can imagine that for "unified") or removed from the glossary (the meaning of "simple" really is common sense).

docs/docs/glossary.md
m-mohr: This is indeed true. For now, I moved the text to the processes, where it belonged to originally. We might want to move the text to another part of the docs later as this is to detailed for a glossary.

docs/docs/glossary.md
m-mohr: Moved to issue #26 for later handling.

docs/docs/glossary.md
m-mohr: Moved to issue #26 for later handling.

docs/docs/glossary.md
m-mohr: Moved to issue #26 for later handling.

PR #48, opened 2018-02-26
Add UDFs to glossary

christophfriedrich:
Mentioning UDFs in the glossary is the only real change.
> If terms as common as "API" are included in the glossary, the not so common abbreviation "UDF" should be explained there too. In general the concept of UDFs is worthy of a brief description there IMO - I tried to come up with something, please replace or amend as you see fit.

All other changes are just fixed typos etc.


docs/docs/processgraphs.md
```diff
@@ -74,7 +74,7 @@ An argument set for a process is defined as follows:
 }
 ```
 
-Whereas a key (`<Key>`) can be any valid JSON key and a value is defined as:
+Where a key (`<Key>`) can be any valid JSON key and a value is defined as:
```
m-mohr: Das klingt falsch...

PR #112, opened 2018-08-16
0.3.0

AlenaDostalova:
I updated glossary and index pages. General idea is that: general information (exmplanation of openEO and words 'simple' and 'unified' were moved on the introduction page, as these are more about the concept of the project and not the technical glossary. In the glossary, I cleaned and simplified the text a bit. In my opinion, also the part 'aggregation and resampling' is too detailed, especially in case, that this text is primarily intended for EO experts.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2,7 +2,14 @@
 
 _Work in progress, please contribute by adding [issues](https://github.com/Open-EO/openeo-api/issues)._
 
-openEO develops an open API that connects clients like R, Python and JavaScript to big Earth observation cloud back-ends in a simple and unified way.
+openEO develops an open application programming interface(API) that connects clients like R, Python and JavaScript to big Earth observation cloud back-ends in a simple and unified way. The acronym openEO contracts two concepts:
+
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -2,7 +2,14 @@
 
 _Work in progress, please contribute by adding [issues](https://github.com/Open-EO/openeo-api/issues)._
 
-openEO develops an open API that connects clients like R, Python and JavaScript to big Earth observation cloud back-ends in a simple and unified way.
+openEO develops an open application programming interface(API) that connects clients like R, Python and JavaScript to big Earth observation cloud back-ends in a simple and unified way. The acronym openEO contracts two concepts:
+
+-	**open**: used here in the context of open source software; open source software is available in source code form, and can be freely modified and redistributed; the openEO project will create open source software, reusable under a liberal open source license (Apache 2.0)
+-	**EO**: Earth observation
+
+Joinly, the openEO targets the processing and analysis of Earth observation data. The main objectives of the project are the following concepts:
```
pramitghosh: #112 
I think there is a typo here. Maybe you meant "Jointly..." @AlenaDostalova ?

PR #156, opened 2018-11-27
Update guidelines-clients.md

bgoesswe:
Fixed Python example syntax and namings


m-mohr:
Good changes, thanks, but PRs need to be done against the 0.4.0 branch.

m-mohr:
I added the changes in commit 215d4b9. Problem with this PR is that it was started using master, which would introduce problems with the branches later. Therefore I included the changes separately. the alternative would have been a completely new PR on a new branch.

PR #193, opened 2019-06-26
process_node has "description" field instead of "process_description"

soxofaan:
minor fix in documentation and openapi.json example:
process_node has "description" field instead of "process_description"


m-mohr:
The intended behavior here is `description`, not `process_description`. This is a leftover from outdated examples. I'll close this PR, but replaced all occurances of `process_description` with `description`. Thanks for the heads-up though!

PR #194, opened 2019-06-27
Pytest based tests to check examples in openapi.json against corresponding schema

soxofaan:
inspired by #193:
- I created pytest based test suite ("suite" is big word for a single but parametrized test function) that checks all the examples in openapi.json against their corresponding schema
- found a couple of issues and fixed them
- added pytest run to travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,10 @@ install:
 - pip install pygments
 - npm install -g asyncapi-generator@0.6.7 # 0.7 only supports AsyncAPI 2.0
 - npm install -g concat-json-files
+- pip install -r tests/requirements.txt
 
 script:
+- pytest tests
 - concat-json-files "processes/*.json" -t "processes.json"
 - mv CHANGELOG.md docs/changelog.md
 - ag subscriptions.json markdown
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,10 @@ install:
 - pip install pygments
 - npm install -g asyncapi-generator@0.6.7 # 0.7 only supports AsyncAPI 2.0
 - npm install -g concat-json-files
-- pip install -r tests/requirements.txt
+- pip install -r tests/requirements.txt  # DIY example validation (PR#194)
 
 script:
-- pytest tests
+- pytest tests  # DIY example validation (PR#194)
 - concat-json-files "processes/*.json" -t "processes.json"
 - mv CHANGELOG.md docs/changelog.md
 - ag subscriptions.json markdown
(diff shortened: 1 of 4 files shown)


openapi.json
```diff
@@ -1685,66 +1685,66 @@
...
+                            "Rcpp": {
+                              "version": "1.0.10",
+                              "links": [
+                                {
+                                  "href": "https://cran.r-project.org/web/packages/Rcpp/index.html"
+                                }
+                              ]
+                            },
+                            "rmarkdown": {
+                              "version": "1.7",
```
m-mohr: I guess this was wrong due to the fact that I usually edit the files in YAML and then convert them to JSON. This conversion probably leads to a number instead of a string. So to ensure it stays a string we should change it to "1.7.0". Could you make this change?

openapi.json
```diff
@@ -1685,66 +1685,66 @@
...
+                            "Rcpp": {
+                              "version": "1.2.0",
+                              "links": [
+                                {
+                                  "href": "https://cran.r-project.org/web/packages/Rcpp/index.html"
+                                }
+                              ]
+                            },
+                            "rmarkdown": {
+                              "version": "1.7",
```
m-mohr: same here with "1.7.0"...

.travis.yml
```diff
@@ -21,8 +21,10 @@ install:
 - pip install pygments
 - npm install -g asyncapi-generator@0.6.7 # 0.7 only supports AsyncAPI 2.0
 - npm install -g concat-json-files
+- pip install -r tests/requirements.txt
 
 script:
+- pytest tests
```
m-mohr: Thanks for adding a solution for the next time. In the long run I hope this is automatically checked by speccy, see https://github.com/wework/speccy/issues/287

openapi.json
```diff
@@ -4762,15 +4762,6 @@
           "example": {
             "description": "A free-form property to include an example for this argument. To represent examples that cannot be naturally represented in JSON, a string value can be used to contain the example with escaping where necessary."
           }
-        },
```
m-mohr: Having an example would be good nevertheless. 

Could you add something like the following snippet, please?

```json
{
  "description": "A percentage between 0 and 100.",
  "required": true,
  "minimum": 0,
  "maximum": 100
}
```

openapi.json
soxofaan: ok will do

openapi.json
soxofaan: ok will do

.travis.yml
soxofaan: I'll add a comment so it can be removed when speccy provides this by default

m-mohr:
Thank you!

PR #222, opened 2019-10-02
errors.json: add placeholders in message of CollectionNotFound and ServiceNotFound

soxofaan:
errors.json: add placeholders in message of CollectionNotFound and ServiceNotFound

### DIFF
--- a/errors.json
+++ b/errors.json
@@ -103,7 +103,7 @@
 	},
 	"CollectionNotFound": {
 		"description": "The requested collection does not exist.",
-		"message": "Collection does not exist.",
+		"message": "Collection '{identifier}' does not exist.",
 		"http": 404,
 		"tags": [
 			"EO Data Discovery",
@@ -130,7 +130,7 @@
 	},
 	"FileNotFound": {
 		"description": "The requested file does not exist.",
-		"message": "File does not exist.",
+		"message": "File '{file}' does not exist.",
 		"http": 404,
 		"tags": [
 			"File Management",
@@ -197,7 +197,7 @@
 	},
 	"ProcessGraphNotFound": {
 		"description": "The requested process graph does not exist.",
-		"message": "Process graph does not exist.",
+		"message": "Process graph '{identifier}' does not exist.",
 		"http": 404,
 		"tags": [
 			"Process Graph Management"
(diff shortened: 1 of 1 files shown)


m-mohr:
Thanks for taking the time to create a PR. I'm not sure about this PR though as not all clients may actually show IDs, but for example a title instead. Also, the changes were only applies to collections and services, but not jobs, process graphs and files etc.

soxofaan:
> as not all clients may actually show IDs, but for example a title instead.

How do you mean: clients *do not want to* or *can not* show IDs? If clients want to show predefined error titles without surprises, shouldn't they use the openEO error code (and map that to a title)  instead of the message anyway?
As a developer I typically have more appreciation for error messages that come with additional usable information (e.g. the id or name of the thing that is not found).

> Also, the changes were only made for collections and services, but not jobs, process graphs and files etc.

I can easily add these, unless this PR is futile :-)

m-mohr:
I meant that it could confuse users if IDs in error messages are shown, but the ID is not shown in general in the UI (e.g. Web Editor). But I guess that's not a big issue and could be solved as mentioned by you.

If it's consistent, I'd merge it. Maybe remove the `_id` from the variables to make it more general-purpose.

soxofaan:
> Maybe remove the _id from the variables to make it more general-purpose.

I used `_id` suffix to be consistent with the API endpoints that typically correspond  with these errors:

-  `GET /collections/{collection_id}`
-  `GET /services/{service_id}`

same issue actually with 
- `GET /jobs/{job_id}`
- `GET /process_graphs/{process_graph_id}`


FYI: errors.json already has another with this suffix: `variable_id`

I'm OK with either way (with or without suffix)

m-mohr:
variable_id is used as there's nothing else in variables that you could really show. For collections, back-ends may choose to show other information than the ID (e.g. title). Usually it should be the ID, but doesn't necessarily need to be. That's why I think a general-purpose {collection} would be better than {collection_id}.

m-mohr:
That's a valid point although I'd hope nobody is replacing it with a full process graph. I just thought to replace them all simply with {identifier}. Except for {file}, which is already used in other places, too. But in the end, this really is bike-shedding. ;-)

Issue #218, opened 2019-09-26
user_id in file management paths


soxofaan:
The file management paths have user id in them:
https://open-eo.github.io/openeo-api/apireference/#tag/File-Management
- GET /files/{user_id}
- GET /files/{user_id}/{path}
- PUT /files/{user_id}/{path}
- DELETE /files/{user_id}/{path}

Why is that, given that the requests already require a bearer authorization header which should identify the user?

flahn:
You are right. We have introduced the `user_id` as argument to enable sharing of data in the workspace. However this would only require `GET /files/{user_id}/{path}` to use the path parameter. And since we get URLs as results from the service and result endpoints that are shareable, we should consider dropping the `user_id` as a parameter in this case.
But maybe I'm missing something.

m-mohr:
Indeed, the user_id in the path is meant mostly for sharing purposes in the future. This is because path is not unique (there could be two users uploading and sharing a test.txt). All other resources have IDs which *should* be unique across all users.

If this answers is sufficient, please close the issue. Otherwise let me know whether you like the parameter to be removed and why. Thank you.

soxofaan:
ok, I understand.
I guess it is recommended for a client to use the authenticated user_id as default? Maybe this could be mentioned in the doc?


On the other hand (and maybe I am picking up old discussions that are a closed deal): as a user I don't expect to see files from other users, let alone other users being able to see my files. 
If you look at Google Drive, Dropbox, S3, etc for example,  users have their own personal "namespace" of paths, which they can not escape by just going up the folder hierarchy. File sharing in these systems also typically happens though a separate path/uri scheme.
But again, maybe I am not aware of reasons to make file sharing as simple as possible.

m-mohr:
> I guess it is recommended for a client to use the authenticated user_id as default? 

Yes.

I'm open to removing the user id. It also depends on the decision we make for #135, I guess...

soxofaan:
I'm in favour of dropping it

m-mohr:
Discussed on the dev telco today. Somewhat agreed on dropping the user_id, but several people (also VITO)  raised that there's not enough experience yet with the /files/... endpoints yet. So there's a preference of dropping it, but before making this change maybe we wait for another month or so and check back again then.

PR #225, opened 2019-10-16
Support multiple OIDC providers (#201) and other Auth improvements

### LINKED ISSUE #86: Authorization Scopes
OpenID Connect (and OAuth) as implemented with #74 allow to add authorization scopes to limit access of users. We should think about the scopes we support and pre-define, e.g.
* read:services
* write:services
* read:jobs
* write:jobs
* ...

### LINKED ISSUE #201: Support multiple OIDC providers + config
Currently, the endpoint for OpenID Connect `GET /credentials/oidc` redirects to a well-known document with the OIDC metadata. It is not supported for back-ends to support multiple authentication providers, which is limiting. Also, for a good UX it may help to add additional information (especially a name) to the response. 

Instead of a redirect, we could respond with a JSON document like this:
```json
{
  "providers": [
    {
      "authority": "https://accounts.google.com",
      "name": "Google"
    },
    {
      "authority": "https://login.eurac.edu",
      "name": "EURAC"
    },
    {
      "authority": "https://sso.uni-muenster.de",
      "name": "WWU Münster SSO"
    }
  ]
}
```

`/.well-known/openid-configuration` has to be appended to the authority.

We may also need to send more information with it, e.g. config parameters for claims or so. Until now, the Web Editor for example assumes that the user id is in the "sub" claim. Where to get the user id and (optionally) user name from? How to communicate user roles?

Examples for options to consider can be viewed here: https://github.com/IdentityModel/oidc-client-js/wiki#required-settings

Interesting to add could be client_id (not sure), response_type and scope.

What to send as Bearer Token? access_token (response_type = 'token) or id_token (response_type = 'id_token')?

### LINKED ISSUE #219: What to send as API bearer token: access token or id token?
This question was raised in #201 but I think it deserves its own ticket:

> What to send as Bearer Token? access_token (response_type = 'token) or id_token (response_type = 'id_token')?

m-mohr:
The new OpenID Connect endpoint, which allows multiple identity providers (see #201). We may add configuration options there, if needed.

It will also clarify the token issue discussed #219. Currently, this PR mandates using the access token, but we may change it to id token during the telco.

This PR is in preparation for the telco on Friday.

Closes #201, closes #219, closes #86.


m-mohr:
> I'm wondering if we should add some kind of flag highlighting whether the Identity Provider allows users to self-register an application (get their client_id/client_secret) or alternatively who they need to contact for this.

Sounds like a good idea, but let me check first whether there's something in the OIDC spec about it.

> Probably 'url' is missing in this sentence:

Thanks, fixed. And also streamlined the Basic Auth process with OIDC.

m-mohr:
@lforesta There's [OpenID Connect Dynamic Client Registration](https://openid.net/specs/openid-connect-registration-1_0.html), which seems to allow issuing client_ids etc. The registration endpoint can be propagated using the [OpenID Well Known Discovery document](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) by setting the `registration_endpoint` property. So I think we could rely on this property?

Anyway, I thought we could still add optional description and links properties, so that IPs can add a helpful description that clients can show and further links a user could follow. What do you think?

openapi.json
```diff
@@ -1770,22 +1770,71 @@
...
... client libraries) is the URL of the identity provider, which conforms to a set of rules:\n1. After appending `/.well-known/openid-configuration` to the URL, a [HTTP/1.1 GET request](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest) to the concatenated URL must return a [OpenID Discovery Configuration Response](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse). The response provides all information required to authenticate using OpenID Connect.\n2. The URL MUST NOT contain a terminating forward slash `/`.\n\nIt is highly RECOMMENDED to implement OpenID Connect for public services in favor of Basic authentication.\n\n**Important:** openEO clients MUST send the access token to back-ends as Bearer token for authorization in subsequent API calls. After authentication, Clients SHOULD request further details about the user (e.g. the user id) from `GET /me`. Back-ends MAY request user information ([]including Claims](https://openid.net/specs/openid-connect-core-1_0.html#Claims)) from the [OpenID Connect Userinfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) using the access token. ",
```
soxofaan: The OpenID Connect RFC and documentation uses the term "OpenID Provider" and abbreviation "OP", instead of "Identity provider". I think these official terms should at least be mentioned for clarity and to avoid confusion.

openapi.json
```diff
@@ -1802,7 +1851,7 @@
     "/credentials/basic": {
       "get": {
         "summary": "HTTP Basic authentication",
-        "description": "This request checks whether the credentials provided in the HTTP Basic header are valid. Returns the user ID and a bearer token for authorization in subsequent API calls. It is RECOMMENDED to implement this authentication method for non-public services only.",
+        "description": "This request checks whether the credentials provided in the HTTP Basic header are valid. Returns an access token to be used as Bearer token for authorization in subsequent API calls.\n\nIt is RECOMMENDED to implement this authentication method for non-public services only.\n\nAfter authentication, Clients SHOULD request further details about the user (e.g. the user id) from `GET /me`.",
```
soxofaan: > whether the credentials provided in the HTTP Basic header are valid

whether valid credentials are provided through Basic HTTP Authentication (HTTP header "Authorization", with type "Basic").

soxofaan: > After authentication, Clients SHOULD request further details

Why *should* clients do that? Currently only the `user_id` field of `/me` endpoint is required, which is typically not that useful for the end user.
I would just write "client can request further ..."

m-mohr: This is related to #218. The user_id is required for the /files/... endpoints. If we remove the user_id from the /files/... endpoints, this can surely be a MAY or so. Otherwise I should explain why it *should* be requested.

openapi.json
m-mohr: Thanks, mostly replaced Identity Provider with OpenID Provider.

openapi.json
m-mohr: I added an explanation. If we agree on #218, I'll remove the corresponding parts from the API.

m-mohr:
I updated the PR with what we have discussed today. Please review it carefully.

openapi.json
```diff
@@ -1797,7 +1871,7 @@
     "/credentials/basic": {
       "get": {
         "summary": "HTTP Basic authentication",
-        "description": "This request checks whether the credentials provided in the HTTP Basic header are valid. Returns the user ID and a bearer token for authorization in subsequent API calls. It is RECOMMENDED to implement this authentication method for non-public services only.",
+        "description": "This request checks whether valid credentials are provided through HTTP Basic Authentication (HTTP header `Authorization`, with type `Basic`). Returns an access token to be used as Bearer token for authorization in subsequent API calls.\n\nIt is RECOMMENDED to implement this authentication method for non-public services only.\n\nAfter authentication, Clients SHOULD request further details about the user from `GET /me` so that they can make requests to the `/files/{user_id}` endpoints.",
```
soxofaan: Now that we have the `basic//` prefix, I would clarify the bit about Bearer token a bit, to avoid the confusion whether this endpoint should return the access token with or without this prefix. E.g. something like

> Returns an access token (without `basic//` prefix) to be used in the Bearer token for authorization in subsequent API calls.

m-mohr: I've added a paragraph to each authentication method about the Bearer Token format, see commit https://github.com/Open-EO/openeo-api/pull/225/commits/371511b6b1dda6d2ddc837382063855a6da5bd19 - Is this clear enough?

openapi.json
```diff
@@ -1797,7 +1871,7 @@
     "/credentials/basic": {
       "get": {
         "summary": "HTTP Basic authentication",
-        "description": "This request checks whether the credentials provided in the HTTP Basic header are valid. Returns the user ID and a bearer token for authorization in subsequent API calls. It is RECOMMENDED to implement this authentication method for non-public services only.",
+        "description": "This request checks whether valid credentials are provided through HTTP Basic Authentication (HTTP header `Authorization`, with type `Basic`).\n\nReturns an access token (without `basic//` prefix) to be used in the Bearer token for authorization in subsequent API calls. The Bearer Token sent to protected endpoints MUST be prefixed with `basic//`. The header would look as follows: `Authorization: Bearer basic//TOKEN` (replace `TOKEN` with the actual access token).\n\nIt is RECOMMENDED to implement this authentication method for non-public services only.\n\nAfter authentication, Clients SHOULD request further details about the user from `GET /me` so that they can make requests to the `/files/{user_id}` endpoints.",
```
soxofaan: Maybe to avoid confusion, because you give a more concrete example about the `Authorization: Bearer` header than about the basic header:

> This request checks whether valid credentials are provided through HTTP Basic Authentication (HTTP header `Authorization`, with type `Basic`)

This request checks whether valid credentials are provided through HTTP Basic Authentication (HTTP header `Authorization: Basic ENCODEDCREDENTIALS`) and returns an access token.\n\nThe returned access token (which SHOULD NOT not include the `basic//` prefix) SHOULD be used in the Bearer token for authorization in subsequent API calls. ...

m-mohr: I made another commit with even more clarifications. I hope it is crystal-clear now.

Issue #272, opened 2020-04-21
Error tagging


soxofaan:
Any reason `JobNotFound` is tagged with `Processes`?
https://github.com/Open-EO/openeo-api/blob/12471721d9c74262ae962f7ad6c746ac77aad382/errors.json#L256-L263

I'm also not sure it's necessary to put `FileNotFound` and `FileContentInvalid` under `Processes`
https://github.com/Open-EO/openeo-api/blob/12471721d9c74262ae962f7ad6c746ac77aad382/errors.json#L123-L130
https://github.com/Open-EO/openeo-api/blob/12471721d9c74262ae962f7ad6c746ac77aad382/errors.json#L173-L180

m-mohr:
I realized the errors.json was quite outdated and the last major update was in 0.4 so took the chance to update it, see PR #275. It should incorporate the requested changes.

PR #275, opened 2020-04-22
Updated errors.json

m-mohr:
Solves #272, #273

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changes
 - `errors.json`: The pre-defined error messages have been reworked.  [#272](https://github.com/Open-EO/openeo-api/issues/272), [#273](https://github.com/Open-EO/openeo-api/issues/273)
-    - Added `UnsupportedApiVersion`, `PermissionsInsufficient`, `ProcessGraphIdDoesntMatch` and `PredefinedProcessExists`.
+    - Added `FolderOperationUnsupported`, `UnsupportedApiVersion`, `PermissionsInsufficient`, `ProcessGraphIdDoesntMatch` and `PredefinedProcessExists`.
     - Added variable `reason` to error `FilePathInvalid` and `type` to `ServiceUnsupported`.
     - Replaced the following error messages. The variables in the messages may have changed, too.
         - `ProcessArgumentUnsupported` -> `ProcessParameterUnsupported`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - Added variable `reason` to error `FilePathInvalid` and `type` to `ServiceUnsupported`.
     - Replaced the following error messages. The variables in the messages may have changed, too.
         - `ProcessArgumentUnsupported` -> `ProcessParameterUnsupported`
-        - `ProcessArgumentRequired` -> `ProcessParameterUnsupported`
         - `ProcessArgumentInvalid` -> `ProcessParameterInvalid`
         - `ProcessParameterMissing` and `ProcessArgumentRequired` -> `ProcessParameterRequired`
         - `ServiceArgumentUnsupported` -> `ServiceConfigUnsupported`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /service_types`: Added optional `title` and `description` properties for service types. [#266](https://github.com/Open-EO/openeo-api/issues/266)
 - `GET /file_formats`: Added optional `description` property for file formats. [#266](https://github.com/Open-EO/openeo-api/issues/266)
 
+### Changes
+- `errors.json`: The pre-defined error messages have been reworked.  [#272](https://github.com/Open-EO/openeo-api/issues/272), [#273](https://github.com/Open-EO/openeo-api/issues/273)
+    - Added `UnsupportedApiVersion`, `PermissionsInsufficient`, `ProcessGraphIdDoesntMatch` and `PredefinedProcessExists`.
+    - Added variable `reason` to error `FilePathInvalid` and `type` to `ServiceUnsupported`.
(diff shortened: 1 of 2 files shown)


errors.json
```diff
@@ -39,107 +39,96 @@
...
 		"tags": [
-			"EO Data Discovery",
-			"Processes"
+			"EO Data Discovery"
 		]
 	},
 	"StorageFailure": {
 		"description": "Server couldn't store file(s) due to server-side reasons.",
-		"message": "Unable to store file(s).",
+		"message": "Unable to store files due to a server error. Please try again later.",
```
soxofaan, line 97: > Please try again later

I would not add that. It kind of suggests that a client can just put things in backof-and-retry-loop and it implies some sort of SLA commitment.

errors.json
```diff
@@ -353,8 +318,8 @@
 		]
 	},
 	"TokenInvalid": {
-		"description": "Authorization token invalid or expired.",
-		"message": "Session has expired.",
+		"description": "Authorization token is invalid or has expired.",
+		"message": "Session has expired. Please authenticate again.",
```
soxofaan, line 322: just nitpicking here:
- I think the description is a better message than the message here (expiry is not the only reason a token can be invalid). Also, I wouldn't add "Please authenticate again." 
- Also change error code to "AuthenticationTokenInvalid" (or "AuthTokenInvalid")?

errors.json
```diff
@@ -39,107 +39,96 @@
...
+		"message": "File path is invalid: {reason}",
 		"http": 400,
 		"tags": [
 			"File Management"
 		]
 	},
 	"FileOperationUnsupported": {
 		"description": "The specified path is not a file and the operation is only supported for files. Path is likely a directory.",
-		"message": "Operation is only supported for files.",
+		"message": "Operation is only supported for files, not folders.",
```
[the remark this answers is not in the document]
m-mohr: This was originally meant for this use case:

You upload `folder/file1.txt` and `folder/file2.txt` via `GET /files/folder/file{1,2}.txt` and then a client calls `DELETE /files/folder`, which the API does not allow, because we basically only work on files always. There are no other use cases that can occur from the file endpoints, are there?

errors.json
m-mohr: That's the nature of 500 error codes. It basically says, it's an error on the server side, please try again and this message is for this use case. If it's a client fault, use one of the other "File*" error messages. Or should there be such a general error also with 400 code?

CHANGELOG.md
```diff
@@ -11,6 +11,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /service_types`: Added optional `title` and `description` properties for service types. [#266](https://github.com/Open-EO/openeo-api/issues/266)
 - `GET /file_formats`: Added optional `description` property for file formats. [#266](https://github.com/Open-EO/openeo-api/issues/266)
 
+### Changes
+- `errors.json`: The pre-defined error messages have been reworked.  [#272](https://github.com/Open-EO/openeo-api/issues/272), [#273](https://github.com/Open-EO/openeo-api/issues/273)
+    - Added `UnsupportedApiVersion`, `PermissionsInsufficient`, `ProcessGraphIdDoesntMatch` and `PredefinedProcessExists`.
+    - Added variable `reason` to error `FilePathInvalid` and `type` to `ServiceUnsupported`.
+    - Replaced the following error messages. The variables in the messages may have changed, too.
+        - `ProcessArgumentUnsupported` -> `ProcessParameterUnsupported`
+        - `ProcessArgumentRequired` -> `ProcessParameterUnsupported`
```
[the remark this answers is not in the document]
m-mohr: Indeed, the `ProcessArgumentRequired` is already captured two lines below. Thanks, will be fixed.

errors.json
```diff
@@ -175,165 +164,141 @@
...
-		"description": "The requested user-defined process does not exist.",
+		"description": "The requested user-defined process does not exist. To be used for all endpoints starting with `/process_graphs`.",
 		"message": "User-defined process '{identifier}' does not exist.",
 		"http": 404,
 		"tags": [
-			"Process Graph Management"
+			"User-Defined Processes"
 		]
 	},
 	"ProcessGraphMissing": {
```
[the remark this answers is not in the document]
m-mohr: `ProcessGraphNotFound` us for requests where /process_graphs/{id} has an non-existing id given (that's why the HTTP code is 404).

`ProcessGraphMissing` is for the case where you don't specify the `process` property in the JSON body of POST/PUT requests to /jobs, /services etc. It could indeed be captured by `ProcessParameterInvalid`. Leaving this open to the implementer and leaving the old message in for backward-compatibility.

errors.json
m-mohr: > * I think the description is a better message than the message here (expiry is not the only reason a token can be invalid).

Well, from a user perspective it's probably hard to understand what invalid means. But I changed it to `Authorization token has expired or is invalid. Please authenticate again.` so that expires is the first reason to read.

> Also, I wouldn't add "Please authenticate again."

I think as a user I would want to know what to do next. Why should that part be a problem? Maybe it's easier to say "log in" though.

> * Also change error code to "AuthenticationTokenInvalid" (or "AuthTokenInvalid")?

As it's really mostly for internal use (the error code name), Ieaving as is for now to not force too many changes.

errors.json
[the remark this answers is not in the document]
soxofaan: > There are no other use cases
probably from the point of of view of the abstract API, but a backend might want to use that error code for other  practical file system related issues as well.
Especially because the code "FileOperationUnsupported" sounds very generic

errors.json
soxofaan: I just mean that the "please try again later" will probably shown to the user in some form and it suggests that the problem will go away magically. I think it's better not to imply that, so that the user takes more appropriate action (e.g. contact support) instead just blindly retrying.
But again, this is nitpicking :)

soxofaan:
Apart from some minor comments I made above, I'm fine with merging this already

errors.json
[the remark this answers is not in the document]
m-mohr: I added a separate `FolderOperationUnsupported` error and changed `FileOperationUnsupported` to be a general-purpose error for file operations.

errors.json
m-mohr: Maybe replace "please try again" with "please try again or contact support?"

PR #277, opened 2020-04-27
Stac spec version tag

soxofaan:
fix broken https://github.com/radiantearth/stac-spec/tree/v0.9.0/ links


m-mohr:
I merged your commit with an additional fix here: https://github.com/Open-EO/openeo-api/commit/401cb4668b7f8b2f63d0e3e51c2b5eb0d925181d

Due to the additional change, this PR doesn't get flagged as merged, but the changes have actually been merged. So closing the PR due to this.

PR #280, opened 2020-05-12
Documentation tweaks related to user defined processes and process graph parameters

soxofaan:
some documentation tweaking suggestions

feel free to cherry pick

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -325,7 +325,7 @@ info:
     ### Accessing process parameters
 
     **Example:** You want to iterate over an array and calculate the absolute value of each value in the array.
-    This can be achieved by executing the `apply` process in openEO and pass a process graph containing a single process `absolute`.
+    This can be achieved by executing the `array_apply` process in openEO and pass a "child" process graph containing a single process `absolute`.
 
     The parameters made available by the "parent" process (`apply` working on the array in this example) to the "child" process graph (the one that calculates of the absolute value of a given array item) are the *process graph parameters*.
     Processes can access parameters by passing a certain type of object as argument to a processes.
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -317,14 +319,15 @@ info:
...
     ```
     
-    This tells the back-end that the process expects the result (i.e. the return value) from another node to be passed as argument. The `<ProcessNodeIdentifier>` is strictly scoped and can only referenced from within the same process graph, i.e. can not be referenced in child or parent process graphs.
+    This tells the back-end that the process expects the result (i.e. the return value) from another node to be passed as argument.
+    The `<ProcessNodeIdentifier>` is strictly scoped and can only reference nodes from within the same process graph, not child or parent process graphs.
 
     ### Accessing process parameters
 
     **Example:** You want to iterate over an array and calculate the absolute value of each value in the array.
     This can be achieved by executing the `apply` process in openEO and pass a process graph containing a single process `absolute`.
```
m-mohr, line 328:
```suggestion
    This can be achieved by executing the `array_apply` process in openEO and pass a process graph containing a single process `absolute`.
```

openapi.yaml
```diff
@@ -317,14 +319,15 @@ info:
...
 
-    The parameters made available by the "parent" process (`apply` in this example) to the "child" process graph are the *process graph parameters*. 
+    The parameters made available by the "parent" process (`apply` working on the array in this example) to the "child" process graph (the one that calculates of the absolute value of a given array item) are the *process graph parameters*.
```
m-mohr, line 330:
```suggestion
    The parameters made available by the "parent" process (`array_apply` in this example) to the "child" process graph are the *process graph parameters*. [`array_apply`](https://processes.openeo.org/#array_apply) for example passes two parameters `x` (each array value consecutively) and `context` (additional data passed from the user). `x` can then be passed to `absolute`.
```

Issue #281, opened 2020-05-20
Clarify that back-ends without billing info are not always free


m-mohr:
Originated inhttps://github.com/Open-EO/openeo-hub/issues/68

The API spec says in /capabilities: 

> Billing related data [...] MUST be specified if the back-end uses any billing related functionalities

We should clarify that this doesn't mean the back-end is free to use. It should say that if you want to implement any billing related functions (e.g. /estimate or plan selection in /jobs) you need to specify the billing info. That doesn't mean you can't bill users at all. The API can be part of a broader payment model where individual API requests don't lead to individual costs, but be part of a "membership fee" model, for example. There you pay indirectly, let's say monthly 1000$ and can use the API with some limits, but without direct costs for requests. So API requests are not directly billed, but it's far from being free to use for all.

christophfriedrich:
Thanks for the clarification.

> Providers may choose to
                      bill users outside of the API, e.g. with a monthly fee
                      that is giving them access a managed virtual machine 
                      running a protected API instance that is not billing based
                      on individual API interactions.

There's a "to" missing after "access".

And IMO the last three lines could be dropped altogether (unnecessary detail).

m-mohr:
Good catch, I shortened to:

> e.g. with a monthly fee that is not depending on individual API interactions.

christophfriedrich:
Elegant, I very much like your shortened version.

Issue #211, opened 2019-09-06
Visualization examples


m-mohr:
In the recent workshop, people first tried to visalize data, but that wasn't very trvilal. Therefore we should ensure there's an easy way to visualize data. Multiple ideas come to mind:
1. Have an visualization example in the collection metadata
2. Have something like a view_collection process that can simply visualize data with some parameters to set.
3. Implement load_collection and save_result in a way that it can visualize without any further processes
4. ...?

sinergise-anze:
Since part of every visualization should be a legend, I would like to suggest another idea: 

4.  it would be nice if the system could generate legends (images) directly from visualization metadata (that is: how the values translate to colors).

Not sure if it helps, but for examples on how legends are defined in Sentinel Hub see [this page](https://www.sentinel-hub.com/faq/how-define-custom-legend).

lforesta:
are we talking about visualization of all data in the cube_view specified in the process graph? This may be a lot, but maybe a back-end could retunr small previews with coarser resolution perhaps?

m-mohr:
Nothing specifically. We are free to choose the best path to go. We could link to previews and/or have visualization defaults in the metadata that save_result can use and/or ...?

m-mohr:
What I'll do here is to allow collection-level assets from STAC. This allows us to link to thumbnails or visualization process graphs from the collections.

Issue #185, opened 2019-06-12
Update STAC


m-mohr:
Update Data Discovery to latest specification version of STAC. STAC 1.0 is expected in Dec. 2019.

m-mohr:
I just allowed any future 1.x version in the spec. So this shouldn't be a big issue anymore. Moving to future to potentially remove 0.9 at some point.

Issue #131, opened 2018-10-08
STAC for job results


m-mohr:
We are currently using our own JSON format and metalink XML for offering job results. As we use STAC already for data discovery, it would be nice to somehow integrate STAC (and WFS) for job result downloading. Could be either an addition/extension or replace our own JSON format.

m-mohr:
It could be as simple as just changing the JSON to follow the STAC Catalog specification (with optional fields from STAC Collections so that it is possible to "upgrade" to a STAC Collection?)

m-mohr:
Another possibility is to output a Single File STAC: https://github.com/radiantearth/stac-spec/tree/master/extensions/single-file-stac

m-mohr:
In the features/stac-for-job-results I have a first draft for a STAC encoding. I simply used STAC Items. Merging would still require a CHANGELOG entry. I need to decide whether that encoding is added on top or replaces the current encoding.

m-mohr:
Dev telco disussions: Providers have geometry and bbox usually available to be written into the job result.

My conclusion: We'll make it GeoJSON compliant and whenever a geometry and bbox is provided, it is also a valid STAC Item.

m-mohr:
Related issue in STAC https://github.com/radiantearth/stac-spec/issues/613

m-mohr:
geometry and bbox is required, datetime can be set to null (which is an invalid STAC at the moment). We'll see how to proceed. May bring that issue back to STAC and see how we solve it.

m-mohr:
STAC 1.0.0-beta will improve the situation a bit. We are still on 0.9, but in docs I already try to pave the way for STAC 1.0.

PR #282, opened 2020-05-22
Several clarifications

m-mohr:
Several clarifications: #281, #131, #211, #185

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -451,7 +451,7 @@ tags:
       
       For data discovery of Earth Observation Collections at the back-ends, openEO strives for compatibility with the specifications [SpatioTemporal Asset Catalog (STAC)](https://stacspec.org/) and [OGC API - Features - Part 1: Core](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html) as far as possible. Implementing the data discovery endpoints of openEO should also produce valid STAC 0.9.0 and OGC API - Features 1.0 responses, including ([partial](#provide-data-for-download)) compatibility with their APIs. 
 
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -1497,6 +1506,9 @@ paths:
                             Domain-specific fields for domains such as EO, SAR and point clouds.
                           * **Custom Properties**:
                             It is generally allowed to add custom fields.
+
+                          This property REQUIRES to add `datacube` to the list
+                          of `stac_extensions`.
```
soxofaan, line 1511: Do I understand this correctly:
as `summaries` is a required field, `datacube` is consequently a required item under `stac_extensions` (a field that is not required)? 
If so: shouldn't `stac_extesions` also be required with at least a `datacube` item?

openapi.yaml
```diff
@@ -3841,6 +3868,13 @@ components:
               title: Thumbnail
               roles:
                 - thumbnail
+            process.json:
+              href: 'https://openeo.org/api/download/583fba8b2ce583fba8b2ce/process.json'
```
soxofaan, line 3872: just a general thought: maybe use "fictional" domain like  `example.openeo.org` instead of a real domain in all examples?

openapi.yaml
m-mohr: Ouh, good catch. I added this to the wrong description. Will move to `cube:dimensions`.

openapi.yaml
m-mohr: Good idea...

m-mohr:
Ouch, I need to wake up. I pushed the review changes to the wrong branch and still merged this. Anyway, the changes are in the draft branch now.

PR #283, opened 2020-05-22
Collection-level assets

m-mohr:
This PR adds collection-level assets as optional property to data discovery, allowing to specify thumbnails or process graphs for visualizations as discussed in issue #211. Fully backward-compatible to RC2 as it is an optional addition, which was already possible before through STAC extensions. I just make it explicit here.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1540,9 +1540,22 @@ paths:
                                   anyOf:
                                     - type: string
                                     - type: number
+                      assets:
+                        type: object
+                        title: Assets
+                        description: >-
+                          Dictionary of asset objects for data that can be downloaded,
+                          each with a unique key.
+                          The keys MAY be used by clients as file names.
+
+                          Implementing this property REQUIRES to add `collection-assets`
+                          to the list of `stac_extensions`.
+                        additionalProperties:
+                          $ref: '#/components/schemas/asset'
                 example:
                   stac_version: 0.9.0
                   stac_extensions:
+                   - collection-assets
                    - datacube
                    - scientific
                    - version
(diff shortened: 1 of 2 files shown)


m-mohr:
Ready for review, #282 has been merged.

m-mohr:
#286 links to examples differently, should I align? 🤔

openapi.yaml
```diff
@@ -4269,19 +4304,21 @@ components:
             2. `license`: A link to the license(s) should be specified if the `license`
             field is set to `proprietary` or `various`.
 
-            3. `derived_from`: Allows linking to the data this collection was derived from.
+            3. `example`: Links to examples of user-defined processes that use this collection.
```
m-mohr, line 4307:
```suggestion
            3. `example`: Links to examples of processes that use this collection.
```

openapi.yaml
```diff
@@ -4027,6 +4019,49 @@ components:
...
+        roles:
+          type: array
+          items:
+            type: string
+          description: |-
+            Purposes of the asset. Can be any value, but commonly used values are:
+            
+            * `thumbnail`: A visualization of the data, usually a lower-resolution true color image in JPEG or PNG format.
+            * `reproducibility`: Information how the data was produced and/or can be reproduced, e.g. the process graph used to compute the data in JSON format.
+            * `data`: The computed data in the format given by the user in the user-defined process.
```
soxofaan, line 4061: maybe I'm missing something, but how are (collection-level) assets linked to user-defined processes?

m-mohr: This OpenAPI schema definition for assets is used in GET /collections/:id (collection assets) and GET /jobs/:id/results (item assets). It doesn't make much sense for collections, but it makes sense in the results where you link to the processed data. It's always hard to decide whether it makes sense to duplicate a lot of schema code just for changing the description a bit... :-|

m-mohr:
```suggestion
            * `data`: The computed data in the format given by the user in the user-defined process (applicable in `GET /jobs/{job_id}/results` only).
```

soxofaan: I understand, but even for `/jobs`: that's not tied to user-defined processes either, right?

m-mohr: Everything you pass in the body for `process` is basically a user-defined process. They don't necessarily need to be stored at /process_graphs. :-) See https://open-eo.github.io/openeo-api/draft/index.html#section/Processes

soxofaan: then I would rather talk of `in the format specified in the process graph` (sounds more concrete to me) instead of using "user defined processes" (which has a more abstract ring to it)

m-mohr: Although I feel like I should say `user-defined process`, similar phrases in the API also say `process graph` at the moment, so changing it to `process graph` until I may change all occurrences.

PR #284, opened 2020-05-25
User-defined process parameters are recursively defined

### LINKED ISSUE #268: User-defined process parameters are recursively defined
The user-defined process parameters are recursively defined, which is not correct. It shouldn't be recursive.

image

See OpenAPI doc -> /processes -> processes -> parameters -> schema -> parameters

Also, return values and service type parameters show the "parameters" json schema keyword, which doesn't make sense there.

m-mohr:
Attempt to fix #268.


m-mohr:
Seems to work fine in Swagger Editor, so merging and waiting for a fix in ReDoc.

Issue #285, opened 2020-05-25
Process Examples incomplete


m-mohr:
Process examples can consist of either arguments only or a process graph at the moment. For process graphs, I probably have forgotten to update the process examples structure when doing the changes for #161. Looking through the processes, the process_graph variant of the examples is rarely used yet (see also https://github.com/Open-EO/openeo-processes/issues/136 ). Another issue is that adding multiple "process graph" examples to a process will increase the size of the GET /results response by far. Also many examples get duplicated. Thus I propose the following:
- Remove process graph examples from the processes +  API
- Add A rel type "example" to the list and allow linking to examples externally. This decreases the response size and makes re-using examples easier.

Strictly speaking, this is a breaking change, but I think it's very rarely used at the moment (maybe even not used at all) and it makes the spec cleaner.

PR #286, opened 2020-05-25
Link to process examples instead of inlining them. #285

m-mohr:
see #285

I think none of the clients (except some of my JS tooling) relies on this / implements this anyway, right? @flahn @bgoesswe @jdries 

For back-ends this is not really a breaking change.

Corresponding process changes are in https://github.com/Open-EO/openeo-processes/pull/151

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4711,7 +4711,7 @@ components:
             (link to older version) and `successor-version` (link to newer version) can also be used
             to show the relation between versions.
 
-            2. `example`: Used to refer to examples of user-defined processes that use this process.
+            2. `example`: Links to examples of other processes that use this process.
 
             3. `cite-as`: For all DOIs associated with the process, the respective DOI
             links should be added.
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4683,31 +4683,20 @@ components:
           $ref: '#/components/schemas/process_exceptions'
         examples:
           type: array
-          description: >-
-            Examples, may be used for tests. Either `process_graph` or
-            `arguments` must be set, never both.
+          description: Examples, may be used for unit tests.
           items:
             title: Process Example
             type: object
-            oneOf:
-              - title: Process Example with Process Graph
-                required:
-                  - process_graph
-                properties:
-                  process_graph:
-                    $ref: '#/components/schemas/process_graph'
-              - title: Process Example with Arguments
-                required:
-                  - arguments
-                properties:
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -4722,7 +4711,9 @@ components:
             (link to older version) and `successor-version` (link to newer version) can also be used
             to show the relation between versions.
 
-            2. `cite-as`: For all DOIs associated with the process, the respective DOI
+            2. `example`: Used to refer to examples of user-defined processes that use this process.
```
flahn, line 4714: What does the example link as process graph show? Is it an alternative implementation of the process or is it an example how it is used in practice? What about authorization restrictions, when linking to an user defined graph?

m-mohr: > What does the example link as process graph show?

A JSON file with a process in JSON format (i.e. a JSON object with at least the member `process_graph` and its nodes).

> Is it an alternative implementation of the process or is it an example how it is used in practice?

An example.

> What about authorization restrictions, when linking to an user defined graph?

Maybe I should remove the `user-defined` here. It's a process provided by the back-end as an example.

m-mohr:
```suggestion
            2. `example`: Links to examples of other processes that use this process.
```

bgoesswe:
I am using it at the QGIS Plugin, but its not really important and no big deal to adapt this. So its fine with me. As far as I know, the Python client is not using them, but I am not up to date there.

m-mohr:
@bgoesswe What are you using it for in the QGIS plug in?

bgoesswe:
There is a Tab with backend information, where you can explore the processes of the backend and there I added examples (if given) to an Infobox text.

soxofaan:
In python client or VITO backend we don't use the examples programatically (only as inspiration for implementation/testing). So fine for me

m-mohr:
@bgoesswe Could you link me to the corresponsing code file?

bgoesswe:
Oh, I think it is currently removed from the Tab, but I still read the metadata from the processes endpoint [here](https://github.com/Open-EO/openeo-qgis-plugin/blob/d36d5577b127a2238ea5e3844ffd9196a0029ee7/models/models.py#L260).

m-mohr:
@bgoesswe That part of the code is still valid though as there's still the "argument-only" examples, which are like 90+% of the current examples. But if you need to re-implement it anyway, then we are fine :-)

PR #288, opened 2020-06-02
Finetune OpenID Connect scope documentation

soxofaan:
https://openeo.org/documentation/1.0/developers/api/reference.html#operation/authenticate-oidc

under `/credientials/oidc`, the `scopes` fields:

    scopes (required)   A list of OpenID Scopes that the client MUST use.
                        If empty, the default scopes of OpenID Connect MUST be used.

This is bit confusing: it is required on one hand, but can be empty (in which case default scopes must be used)

I'm also not sure what _default_ scopes mean in this context, but as far as I know, only `openid` is a required scope to enable OpenID Connect. I would just simplify the doc to state that.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2063,23 +2063,6 @@ paths:
         Discovery](http://openid.net/specs/openid-connect-discovery-1_0.html).
 
 
-        For each OpenID provider, a name and the issuer location MUST be listed.
-        A description and related links can OPTIONALLY be added. The [issuer
-        location](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig)
-        (also referred to as 'authority' in client libraries) is the URL of the
-        OpenID provider, which conforms to a set of rules:
-
-        1. After appending `/.well-known/openid-configuration` to the URL, a
-        [HTTP/1.1 GET
-        request](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest)
-        to the concatenated URL must return a [OpenID Discovery Configuration
-        Response](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse).
-        The response provides all information required to authenticate using
-        OpenID Connect.
-
-        2. The URL MUST NOT contain a terminating forward slash `/`.
-
-
         It is highly RECOMMENDED to implement OpenID Connect for public services
         in favor of Basic authentication.
 
@@ -2087,8 +2070,8 @@ paths:
         openEO clients MUST use the **access token** as part of the Bearer token
         for authorization in subsequent API calls (see also the information
         about Bearer tokens in this document). Clients MUST NOT use the id token
-        or the authorization code. The access token provided by this the OpenID
-        Provider does not always provide information about the issuer (i.e. the
+        or the authorization code. The access token provided by an OpenID
+        Provider does not necessarily provide information about the issuer (i.e. the
         OpenID provider) and therefore a prefix MUST be added to the Bearer
         Token sent in subsequent API calls to protected endpoints. The Bearer
         Token sent to protected endpoints MUST consist of the authentication
@@ -2129,7 +2112,6 @@ paths:
                       required:
                         - id
                         - issuer
-                        - scopes
                         - title
                       properties:
                         id:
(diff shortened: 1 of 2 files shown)


soxofaan:
I pushed another related commit:
- endpoint doc uses `name`, while field is actually `id`
- `a description ... optionally added`: that's bit redundant given the field listing below that

m-mohr:
I agree with most of this, thanks.

I'm not sure about the scopes. I think the piece to fix here is the scopes requirement as the description also doesn't list the scopes: "For each OpenID provider, a name and the issuer location MUST be listed." It doesn't say scopes and the scopes description basically allows it bein non-existant.

openapi.yaml
```diff
@@ -2063,9 +2063,8 @@ paths:
         Discovery](http://openid.net/specs/openid-connect-discovery-1_0.html).
 
 
-        For each OpenID provider, a name and the issuer location MUST be listed.
-        A description and related links can OPTIONALLY be added. The [issuer
-        location](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig)
+        For each OpenID provider, an id and the issuer location MUST be listed.
+        The [issuer location](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig)
         (also referred to as 'authority' in client libraries) is the URL of the
```
m-mohr, line 2068:
```suggestion
        (also referred to as 'authority' in some client libraries) is the URL of the
```

openapi.yaml
```diff
@@ -2150,11 +2149,8 @@ paths:
                         scopes:
                           type: array
                           description: >-
-                            A list of OpenID Scopes that the client MUST use.
-
-
-                            If empty, the default scopes of OpenID Connect MUST
-                            be used.
+                            A list of OpenID Connect scopes that the client MUST use.
+                            The list MUST at least contain the `openid` scope.
```
m-mohr, line 2153:
```suggestion


                            If no scope is specified, clients MUST use the OpenID Connect
                            default scope `openid`.
```

Please also remove the requirement for scopes here: https://github.com/Open-EO/openeo-api/pull/288/files#diff-fe030a7c1568b3decf599edf399be7f4L2132

openapi.yaml
soxofaan: will do

soxofaan:
already pushed some updates (all to be squashed before merging)

some more notes

> I think the piece to fix here is the scopes requirement as the description also doesn't list the scopes: "For each ...

Indeed, currently `id`, `issuer`, `scopes` and `title` are required, but the (endpoint) description only mentions name/id and issuer as "MUST be listed". I considered fixing this incomplete situation by moving stuff from the endpoint description to the field descriptions, but didn't want to move too much at once. It would make things cleaner and more concise however. What do you think?


> If no scope is specified, clients MUST use the OpenID Connect     default scope `openid`.

This allows a backend to specify a non-empty list without `openid`, which breaks OpenID Connect. I would specify something like "if scopes are specified, the list MUST at least contain the `openid` scope".

openapi.yaml
```diff
@@ -2063,10 +2063,9 @@ paths:
         Discovery](http://openid.net/specs/openid-connect-discovery-1_0.html).
 
 
-        For each OpenID provider, a name and the issuer location MUST be listed.
-        A description and related links can OPTIONALLY be added. The [issuer
-        location](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig)
-        (also referred to as 'authority' in client libraries) is the URL of the
+        For each OpenID provider, an id and the issuer location MUST be listed.
```
m-mohr, line 2066:
```suggestion
        For each OpenID provider, the issuer location with an id and a title MUST be listed.
```

m-mohr:
> already pushed some updates (all to be squashed before merging)

Thanks.


> Indeed, currently `id`, `issuer`, `scopes` and `title` are required, but the (endpoint) description only mentions name/id and issuer as "MUST be listed".

I now realized that name was referring to title. So to be very explicit, we should list id, title and issuer.

> I considered fixing this incomplete situation by moving stuff from the endpoint description to the field descriptions, but didn't want to move too much at once. It would make things cleaner and more concise however. What do you think?

What exactly? The first paragraphs are just an overview and I think the information in there is already part of the fields, isn't it?

> This allows a backend to specify a non-empty list without `openid`, which breaks OpenID Connect. I would specify something like "if scopes are specified, the list MUST at least contain the `openid` scope".

Oh yes, add that, please. I just checked whether we can also enforce this with the schema, but OpenAPI unfortunately doesn't allow that. JSON Schema has the contains keyword, but it's not part of OpenAPI yet. We could add x-contains as reminder for later (JSON Schema is on its way) though: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#specificationExtensions

that would be 
````
x-contains:
  const: "openid"
```

soxofaan:
> What exactly? The first paragraphs are just an overview and I think the information in there is already part of the fields, isn't it?

I just mean that it looks redundant and prone to inconsistencies to have the same information:
- textually in the endpoint description ("for each ... MUST be listed .... [other] ... OPTIONALLY added")
- structurally in the field metadata ("required" vs "not required")

I would propose to move the whole paragraph of the "issuer location" that's currently in the endpoint description to  the description of the `issuer` field. And drop the textual specification of MUST/OPTIONAL fields.

But, I don't feel very strong about this, it's just a minor suggestion to streamline things a bit

m-mohr:
Yes, that works for me. Merge it, please. :-)

soxofaan:
> Oh yes, add that, please. 

It's still like in my original PR, so I'll leave it like that.

> enforce this with the schema, but OpenAPI unfortunately doesn't allow that.  

One could argue that it's not that big of an issue because leaving out `openid` scope would be a violation of the OIDC spec, not necessarily openeo spec. I don't think it's the responsibility of openeo API to enforce a whole other spec.

m-mohr:
> It's still like in my original PR, so I'll leave it like that.

I think your proposed "If scopes are specified, the list MUST at least contain the `openid` scope." is a bit clearer than just "The list MUST at least contain the `openid` scope.". With that it's ready to be merged, I think.

> One could argue that it's not that big of an issue because leaving out `openid` scope would be a violation of the OIDC spec, not necessarily openeo spec. I don't think it's the responsibility of openeo API to enforce a whole other spec.

I think if we have a requirement in text, we should try to enforce them as much as possible in the schemas. Otherwise we could also just leave it out of the text as well and just refer to OpenID Connect. But I'm not strongly feeling about this workaround.

soxofaan:
ok, squashed everything and force-pushed

m-mohr:
Thank you!

PR #290, opened 2020-06-09
added exceptions for ndvi process

bossie:
They match the exceptions specified in the ndvi process description in openeo-processes.

### DIFF
--- a/errors.json
+++ b/errors.json
@@ -407,5 +407,37 @@
 		"tags": [
 			"Secondary Services Management"
 		]
+	},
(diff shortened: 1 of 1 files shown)


m-mohr:
Thanks for opening the PR, but the error list in the API is not meant to contain process specific errors. They are only defined in the processes itself to avoid dependencies between processes and the API (they are meant to be released independently).

m-mohr:
We removed everything process specific for the next release in PR #275 as there were some left-overs...

PR #291, opened 2020-06-12
STAC clarifications

m-mohr:
- Allow all STAC versions in range 0.9.x and 1.x.x for being more future proof. 0.9.0 and 1.x.x are vastly compatible and 1.x.x is not allowed to include braking changes anyway (SemVer followed by STAC).
- Beforehand STAC was not clear how to specify unlocated Items, clarified that now that is has been clarified in STAC: Unlocated results set `geometry` to `null` and omit the `bbox` property (applies only to `GET /jobs/{job_id}/results`).
- Formally forbid 5 elements in bounding boxes.
- Clarify difference between STAC specification and STAC API.
- Clarify that a copy of the STAC Item should be part of the assets in a batch job download.

This is backward-compatible with all back-ends, but may need some subtle changes in clients.

Related to:
* https://github.com/radiantearth/stac-spec/pull/844
* https://github.com/opengeospatial/ogcapi-features/issues/401
* https://github.com/radiantearth/stac-spec/issues/843
* https://github.com/radiantearth/stac-spec/issues/850


CHANGELOG.md
```diff
@@ -75,6 +78,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clarified that back-ends not supporting pagination will return all resources.
 - Clarified how `from_parameter` is resolved in case no value is given.
 - Clarified `GET .../logs` endpoint behaviour.
+- Clarify difference between STAC specification and STAC API.
+- Clarify that a copy of the STAC Item should be part of the assets in a batch job download.
```
soxofaan, line 82: bit of nitpicking: I read the "should" as "MUST", while the spec diff below states "RECOMMENDED".
Maybe change to something like

> Added recommendation to include a link to the STAC Item in assets of batch job download

openapi.yaml
```diff
@@ -4191,9 +4235,10 @@ components:
             $ref: '#/components/schemas/link'
     stac_version:
       type: string
-      description: The STAC version implemented.
-      enum:
-        - 0.9.0
+      description: >-
+        The [version of the STAC specification](https://github.com/radiantearth/stac-spec/releases),
+        which MAY not be equal to the [STAC API version](#section/STAC).
+        Supports versions 0.9.x and 1.x.x.
```
soxofaan, line 4241: just out of curiosity: is there a guarantee that 1.x.x will strictly follow 0.9.x, without a 0.10.x in between?

m-mohr: Only my word as core contributor, but 1.0.0-beta is already released, too ;-)

CHANGELOG.md
m-mohr: should is never a must, but will change:

```suggestion
- Clarify that a copy of the STAC Item is recommended to be part of the assets in a batch job download.
```

soxofaan: yes I know, it's just that kind of terminology is less strictly or explicitly defined in the context of a CHANGELOG listing

m-mohr:
Will merge on Wednesday, if no other comments come up.

m-mohr:
Due to other demands, merging now. Got feedback from @flahn and @soxofaan and most changes (i.e. Item assets) are not relevant for the Hub or will be "fixed" through js-commons or vue-components, so should not be a problem.

PR #295, opened 2020-06-16
Finetune documentation about user defined processes and process parameters

soxofaan:
The size of the diff is larger than what actually changed:

- be more consistent in the order of ResultReference/UserDefinedProcess/ParameterReference . These argument types show up in four places in the docs, and using the same order everywhere makes things more clear

- be more consistent with the `apply/absolute` example. It was now a mix of `array_apply` and `apply`. I also pulled it up into the UserDefinedProcess docs for even more reuse.

- some minor finetuning, title casing and typo fixing

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -281,24 +281,44 @@ info:
 
     ```
     <Arguments> := {
-      "<ParameterName>": <string|number|boolean|null|array|object|UserDefinedProcess|ParameterReference|ResultReference>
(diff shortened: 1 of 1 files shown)


m-mohr:
Thank you, that looks good.
The title casing is not really consistently used throughout the API definition, so I think I'd like to "normal-case" case that (not sure how that's called in reality), especially as "Accessing Results Of Other Process Nodes" looks pretty weird with non-nouns being upper-cased.

soxofaan:
force pushed amended commit without the title casing

m-mohr:
Thank you!

PR #296, opened 2020-06-17
update evi example process graph to 1.0-style math processes

soxofaan:


m-mohr:
That's a good alignment, also with the pg-evi-example.json. I just realized that the PNG is bugged though.

PR #298, opened 2020-06-23
Content Model Simplification

m-mohr:


m-mohr:
Okay, seems to be a no-brainer. :-)

PR #302, opened 2020-07-01
Add type to UDF Runtime for better code generation

m-mohr:
This adds a required field `type` to each UDF runtime in `/udf_runtimes`. This is a breaking change, which means each back-end needs to add the field to be valid against the spec.

Ususally, I'd not like breaking changes in this stage, but recently I worked with @aljacob to improve the results you get from openapi-generator. This change is required to get a well-defined class inheritance for the generated classes, without this change the generated classes are pretty useless. Thus we thought it's an easy fix for back-ends to make the OpenAPI document more future-proof and allow new implementers to auto-generate better code. We hope this eases adoption. 

Due to the breaking change I'm asking for approvals / concerns / vetos from back-ends here. This shouldn't be an issue for clients.


m-mohr:
As all (known) 1.0 APIs supporting UDFs have approved, I'll merge it now.

PR #306, opened 2020-07-03
Finetune pagination_limit description

soxofaan:
- some minor structure/wording tweaks
- dropped the `The only exception ..` part because that seems redundant or even confusing (it reads a bit like the `links` listing should not change across pages, while the browsing links actually SHOULD change on every page)

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -5616,23 +5616,24 @@ components:
       name: limit
       description: |-
         This parameter enables pagination for the endpoint and specifies the maximum number of
-        elements that arrays in the top-level object (e.g. jobs or log entries) are allowed to contain.
-        The only exception is the `links` array, which MUST NOT be paginated as otherwise the
-        pagination links may be missing ins responses.
+        elements that arrays in the top-level object (e.g. collections, processes, batch jobs,
+        secondary services, log entries, etc.) are allowed to contain.
+        The `links` array MUST NOT be paginated like the resources,
+        but instead contain links related to the paginated resources
+        or the pagination itself (e.g. a link to the next page).
         If the parameter is not provided or empty, all elements are returned.
-        
-        Pagination is OPTIONAL and back-ends and clients may not support it.
+
+        Pagination is OPTIONAL: back-ends or clients may not support it.
         Therefore it MUST be implemented in a way that clients not supporting
-        pagination get all resources regardless. Back-ends not supporting 
-        pagination will return all resources.
-        
-        If the response is paginated, the links array MUST be used to propagate the 
-        links for pagination with pre-defined `rel` types. See the links array schema
+        pagination get all resources regardless. Back-ends not supporting
+        pagination MUST return all resources.
+
+        If the response is paginated, the `links` array MUST be used to communicate the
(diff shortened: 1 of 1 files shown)


soxofaan:
and to piggy-back this issue: concerning the batch job listings: is a backend expected to keep record of all batch jobs a user ever submitted? Can't we add a minor disclaimer that a backend is allowd to prune entries of old batch jobs?

openapi.yaml
```diff
@@ -5617,22 +5617,19 @@ components:
...
+        pagination MUST return all resources.
+
+        If the response is paginated, the links array MUST be used to communicate the
+        links for browsing the pagination with pre-defined `rel` types. See the links array schema
         for supported `rel` types.
-        
-        *Note:* Implementations can use all kind of pagination techniques, depending on what is
-        supported best by their infrastructure. So it doesn't care whether it is page-based,
-        offset-based or uses tokens for pagination. The clients will use whatever is specified
+        Backend implementations can use all kind of pagination techniques, depending on what is
```
m-mohr, line 5630: Here's a new line missing

openapi.yaml
```diff
@@ -5617,22 +5617,19 @@ components:
       description: |-
         This parameter enables pagination for the endpoint and specifies the maximum number of
         elements that arrays in the top-level object (e.g. jobs or log entries) are allowed to contain.
-        The only exception is the `links` array, which MUST NOT be paginated as otherwise the
```
m-mohr, line 5620: Why has this been removed? It's a crucial information.

m-mohr:
> is a backend expected to keep record of all batch jobs a user ever submitted? Can't we add a minor disclaimer in the API that a backend is allowed to prune entries of old batch jobs?

That's nothing the API should cater for as the user actually never sees it. That API doesn't forbid such as "auto-delete" behavior. Providers should inform in their terms of service or wherever it's seen by users and then can do whatever he likes.

openapi.yaml
soxofaan: That is intended, because both original paragraphs are about how the `rel` links relate to pagination.

openapi.yaml
soxofaan: Ok I'll try to rewrite it a bit

soxofaan:
> That's nothing the API should cater for as the user actually never sees it.

I'm not sure about that. E.g. client might reuse the description form the API in their documentation, or link to it.

m-mohr:
I hope they don't, at least they are not written for that.

soxofaan:
another bit of piggy-backing:

for logs of batch jobs and secondary services, there is an `offset`  parameter. Does that mean that backend MUST use the offset based pagination? Which conflicts currently with this part of the description of `limit`:
> Implementations can use all kind of pagination techniques, depending on what is supported best by their infrastructure.

m-mohr:
Yes, for logs it's always offset. That's due to the reasons that logs may be "streamed" in real-time.

soxofaan:
force-pushed another update, related to the offset-based pagination for logs

m-mohr:
Thanks!

Issue #305, opened 2020-07-03
Processes: pre-defined + user-defined + sharing: namespacing/prefixing?


soxofaan:
(I remember discussions about this but couldn't find in-depth github ticket about this, feel free to close as duplicate)

At VITO, we're working on support for user-defined processes and basic sharing, which raises some concerns about naming collisions, not only between pre-defined and user-defined processes, but also across users when some kind of sharing mechanism is in place.

Some data points:

- From #256:

    > Decide whether user-defined processes should be prefixed in process graphs (e.g. user:ndvi) - I'll leave this up to the back-end for now. We may revisit this for 1.0 final.

- The description of the `process_id` component:
   https://github.com/Open-EO/openeo-api/blob/12471721d9c74262ae962f7ad6c746ac77aad382/openapi.yaml#L4661-L4663
   I think there is a contradiction/inconsistency here:  `process_id` MUST be unique .... but if backend adds same name, it's ok to be not unique anymore (can user still update their UDP afterwards?). I think this is an indication that it "needs more work".

- Process ids currently must follow this regex: 
    https://github.com/Open-EO/openeo-api/blob/12471721d9c74262ae962f7ad6c746ac77aad382/openapi.yaml#L4667
    meaning Latin characters, digits and underscore. Which means there is very little wiggle room to implement some kind of robust namespace/prefix scheme. Also somewhat relevant here, especially when sharing across users comes in the picture: in context of OpenID Connect, the user id returned by the OIDC provider might not fit `\w+` (dashes, dots, ...)

I don't have a concrete suggestions at the moment, but just wanted to start off this discussion

m-mohr:
> I remember discussions about this but couldn't find in-depth github ticket about this, feel free to close as duplicate [...]
 >   > Decide whether user-defined processes should be prefixed in process graphs (e.g. user:ndvi) - 

I think that was discussed mainly in calls. Result was that some back-ends want prefixes for UDP and some not. Thus, there's now this compromise in the docs:

> Back-ends may choose to enforce a prefix for user-defined processes while storing the process, e.g. user_ndvi with user_ being the prefix. Prefixes must still follow the pattern.

>   I think there is a contradiction/inconsistency here:  `process_id` MUST be unique ....

... from the user-perspective at the time of submission.

> but if backend adds same name, it's ok to be not unique anymore (can user still update their UDP afterwards?). I think this is an indication that it "needs more work".

[...]

PR #309, opened 2020-07-06
#305: Add namespace to process graph nodes

m-mohr:
see #305 for reasons and discussions.

Please review carefully and approve, if this change is fine for you. Otherwise comment here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /.well-known/openeo` and `GET /`: `production` fields default to `false` instead of `true`.
 - `GET /collections` and `GET /collections/{collection_id}`: Additional dimensions in `cube:properties` can only be of type `other`.
 - Allow all STAC versions that are compatible to STAC 0.9.0.
+- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The old behavior to prioritize user-defined over pre-defined processes has been removed. [#305](https://github.com/Open-EO/openeo-api/issues/305)
 - Added `format: commonmark` to all properties supporting CommonMark formatting.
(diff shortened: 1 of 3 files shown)


jdries:
Does this include allowing URI's as process_id? Or will the namespace be a URI?

m-mohr:
No URLs allowed yet. This seems to be too much of a change for 1.0, see also my https://github.com/Open-EO/openeo-api/issues/305#issuecomment-654166844. I think it might be better to leave URIs up to clients for now, but the API explicitly says this might be extended later. In this case it would be URI in namespace, because process_id at the moment only allows the \w+ pattern.

soxofaan:
looks good, thanks for the quick PR

flahn:
just to understand this, the user will specify the namespace when sending an UDP to the back-end, right? If so, won't we need an enumeration for the namespaces that can be used on the back-end? Or ist it simply "user" and "predefined" or is it completely up to the user to chose any namespace?

m-mohr:
@flahn Currently it's only "user" (= user-defined) and null (= pre-defined). There could be more in the future that a user could specify, but for now it's those two. Having that said, those two can easily be set by the clients and user's don't even need to notice the namespaces. When constructing a process graph with a builder, the client knows where the process originates from (/processes or /process_graphs) and then can set the flag accordingly for the user.

flahn:
Right. If it is limited to "user" and null this can be done without user interaction and can stay hidden. However, if it is planned in the future to use other namespaces I can't hide it. Then the user has somehow to be informed which namespace they can specify and what it means.

m-mohr:
Still, only it would only be required to expose this for other namespace than the one listed above. If you read the issue, there are some examples what could be added in the future and yes, those would likely somehow need to be specified by the user, but even those could be somehow hidden that it's not too confusing. For example, the "URL" use-case could just be that a user specifies a URL and then that it's a URL is detected by the client and set accordingly. But that's the future, for now it can be hidden in clients.

m-mohr:
Seems to be consensus => merging.

PR #313, opened 2020-07-13
Revert namespace default

m-mohr:
I just realized again why we haven't had namespaces ( https://github.com/Open-EO/openeo-api/pull/309 ) in the API: User could simply replace missing pre-defined processes (e.g. normalized_difference) with our "alternative" processing instructions in openeo-processes. Just store it as user-defined process and it still runs. That's now broken with namespaces.

Thus, I partly revert the changes introduced in #309: The default behavior for a process graph is again preferring user-defined over pre-defined processes.
The namespaces are now a backward compatible addition to explicitly choosing the pre- or user-defined process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /.well-known/openeo` and `GET /`: `production` fields default to `false` instead of `true`.
 - `GET /collections` and `GET /collections/{collection_id}`: Additional dimensions in `cube:properties` can only be of type `other`.
 - Allow all STAC versions that are compatible to STAC 0.9.0.
-- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The old behavior to prioritize user-defined over pre-defined processes is still the default. [#305](https://github.com/Open-EO/openeo-api/issues/305)
+- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The default behavior still prioritizes user-defined over pre-defined processes. [#305](https://github.com/Open-EO/openeo-api/issues/305)
 - Added `format: commonmark` to all properties supporting CommonMark formatting.
 - `errors.json`: The pre-defined error messages have been reworked.  [#272](https://github.com/Open-EO/openeo-api/issues/272), [#273](https://github.com/Open-EO/openeo-api/issues/273)
     - Added `FolderOperationUnsupported`, `UnsupportedApiVersion`, `PermissionsInsufficient`, `ProcessGraphIdDoesntMatch` and `PredefinedProcessExists`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /.well-known/openeo` and `GET /`: `production` fields default to `false` instead of `true`.
 - `GET /collections` and `GET /collections/{collection_id}`: Additional dimensions in `cube:properties` can only be of type `other`.
 - Allow all STAC versions that are compatible to STAC 0.9.0.
-- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The old behavior to prioritize user-defined over pre-defined processes has been removed. [#305](https://github.com/Open-EO/openeo-api/issues/305)
+- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The old behavior to prioritize user-defined over pre-defined processes is still the default. [#305](https://github.com/Open-EO/openeo-api/issues/305)
 - Added `format: commonmark` to all properties supporting CommonMark formatting.
 - `errors.json`: The pre-defined error messages have been reworked.  [#272](https://github.com/Open-EO/openeo-api/issues/272), [#273](https://github.com/Open-EO/openeo-api/issues/273)
     - Added `FolderOperationUnsupported`, `UnsupportedApiVersion`, `PermissionsInsufficient`, `ProcessGraphIdDoesntMatch` and `PredefinedProcessExists`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /.well-known/openeo` and `GET /`: `production` fields default to `false` instead of `true`.
 - `GET /collections` and `GET /collections/{collection_id}`: Additional dimensions in `cube:properties` can only be of type `other`.
 - Allow all STAC versions that are compatible to STAC 0.9.0.
-- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The default behavior still prioritizes user-defined over pre-defined processes. [#305](https://github.com/Open-EO/openeo-api/issues/305)
+- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The default behavior has not changed. [#305](https://github.com/Open-EO/openeo-api/issues/305)
 - Added `format: commonmark` to all properties supporting CommonMark formatting.
 - `errors.json`: The pre-defined error messages have been reworked.  [#272](https://github.com/Open-EO/openeo-api/issues/272), [#273](https://github.com/Open-EO/openeo-api/issues/273)
     - Added `FolderOperationUnsupported`, `UnsupportedApiVersion`, `PermissionsInsufficient`, `ProcessGraphIdDoesntMatch` and `PredefinedProcessExists`.
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4728,6 +4728,7 @@ components:
         Unique identifier for the process.
         
         MUST be unique across its namespace (e.g. pre-defined processes or user-defined processes).
+        Additionally, user-defined processes with an identifier equal to any pre-defined process MUST be rejected.
       pattern: '^\w+$'
       example: ndvi
     process_summary:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4711,10 +4711,17 @@ components:
       example: null
       description: |-
         The namespace the `process_id` is valid for.
+
+        The following options are pre-defined by the openEO API, but additional
+        namespaces may be introduced by back-ends or in a future version of the API.
         
-        By default, the namespace is `null`, which includes all pre-defined processes listed at `GET /processes`.
-        The namespace `user` refers to all user-defined processes listed as `GET /process_graphs`.
-        Additional namespaces may be introduced by back-ends or in a future version of the API.
+        * `null` (default): Checks both user-defined and pre-defined processes,
+           but prefers user-defined processes if both are available.
+           This allows users to replace missing pre-defined processes with
+           user-defined processes, e.g. processes from https://processes.openeo.org
+           that have a process graph included.
+        * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
+        * `user`: Uses exclusively the user-defined processes listed at  listed as `GET /process_graphs`.  
     process_id:
       type: string
       description: |-
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4731,11 +4731,13 @@ components:
     process_id:
       type: string
       description: |-
-        Unique identifier for the process.
+        An identifier for the process.
         
-        MUST be unique across its namespace (e.g. pre-defined processes or user-defined processes).
-        Clients SHOULD warn the user if a user-defined process is added with an identifier that
-        exists in the pre-defined processes.
+        The identifier MUST be unique across its namespace (e.g. pre-defined
+        processes or user-defined processes).
+
+        Clients SHOULD warn the user if a user-defined process is added with the 
+        same identifier as one of the pre-defined process.
       pattern: '^\w+$'
       example: ndvi
     process_summary:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4717,16 +4717,11 @@ components:
         
         * `null` (default): Checks both user-defined and pre-defined processes,
            but prefers user-defined processes if both are available.
-           This allows users to replace missing pre-defined processes with
-           user-defined processes for portability, e.g. common processes from
-           [processes.openeo.org](https://processes.openeo.org) that have a process graph
-           included.
+           This allows users to add missing pre-defined processes for portability,
+           e.g. common processes from [processes.openeo.org](https://processes.openeo.org)
+           that have a process graph included.
            If a user-defined process is prioritized over a pre-defined process,
            a log entry SHOULD be added.
-           To improve reproducibility, it is RECOMMENDED that back-ends persist
-           the namespace that would be selected at the time the process is
-           submitted so that adding or removing processes doesn't change the
-           processing results in the future.
         * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
         * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
 
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4721,25 +4721,21 @@ components:
            user-defined processes for portability, e.g. common processes from
            [processes.openeo.org](https://processes.openeo.org) that have a process graph
            included.
+           If a user-defined process is prioritized over a pre-defined process,
+           a log entry SHOULD be added.
+           To ensure reproducibility, back-ends SHOULD persist the namespace that
+           is prioritized at the time the process is submitted so that adding or
+           removing processes doesn't change the processing results in the future.
         * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
         * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
-
-        Note on collisions: The only way a collision between user-defined processes and
-        pre-defined processes can occur is when back-ends introduce a new process with an
-        identifier already used by users. User-defined processes with an identifier used by
-        a pre-defined process MUST be rejected. To ensure reproducibility, back-ends SHOULD
-        persist the namespace that is prioritized at the time the process is submitted.
-        If a collision occurs after a has back-end introduced a new pre-defined process
-        (i.e. a user-defined process is prioritized over a pre-defined process),
-        a log entry should be added so that the user is informed about the availability of the
-        new process.
     process_id:
       type: string
       description: |-
         Unique identifier for the process.
         
         MUST be unique across its namespace (e.g. pre-defined processes or user-defined processes).
-        Additionally, user-defined processes with an identifier equal to any pre-defined process MUST be rejected.
+        Clients SHOULD warn the user if a user-defined process is added with an identifier that
+        exists in the pre-defined processes.
       pattern: '^\w+$'
       example: ndvi
     process_summary:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -205,7 +205,7 @@ info:
     
     **Back-end providers** MUST follow the schema for predefined processes as in [`GET /processes`](/#tag/Process-Discovery) to define new processes. This includes:
 
-    * Choosing a intuitive and ideally unique name as process id, consisting of only letters (a-z), numbers and underscores.
+    * Choosing a intuitive process id, consisting of only letters (a-z), numbers and underscores. It MUST be unique across the pre-defined processes.
     * Defining the parameters and their exact (JSON) schemes.
     * Specifying the return value of a process also with a (JSON) schema.
     * Providing examples or compliance tests.
@@ -213,7 +213,7 @@ info:
     
     **Users** MUST follow the schema for user-defined processes as in [`GET /process_graphs`](/#tag/User-Defined-Processes) to define new processes. This includes:
 
-    * Choosing a intuitive and ideally unique name as process id, consisting of only letters (a-z), numbers and underscores.
+    * Choosing a intuitive name as process id, consisting of only letters (a-z), numbers and underscores. It MUST be unique across the user-defined processes.
     * Defining the algorithm as a process graph.
     * Optionally, specifying the additional metadata for processes.
 
@@ -273,7 +273,7 @@ info:
 
     One of the nodes in a map of processes (the final one) MUST have the `result` flag set to `true`, all the other nodes can omit it as the default value is `false`. Having such a node is important as multiple end nodes are possible, but in most use cases it is important to exactly specify the return value to be used by other processes. Each child process graph must also specify a result node similar to the "main" process graph.
 
-    `process_id` MUST be a valid process ID in the `namespace` given. By default, the namespace is `null`, which includes all pre-defined processes listed at `GET /processes`. The namespace `user` refers to all user-defined processes listed as `GET /process_graphs`. Additional namespaces may be introduced later.
+    `process_id` MUST be a valid process ID in the `namespace` given. Clients SHOULD warn the user if a user-defined process is added with the same identifier as one of the pre-defined process.
 
     ### Arguments
 
@@ -4730,10 +4730,8 @@ components:
     process_id:
       type: string
       description: |-
-        An identifier for the process.
-        
-        The identifier MUST be unique across its namespace (e.g. pre-defined
-        processes or user-defined processes).
+        The identifier for the process. It MUST be unique across its namespace
+        (e.g. pre-defined processes or user-defined processes).
 
         Clients SHOULD warn the user if a user-defined process is added with the 
         same identifier as one of the pre-defined process.
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4729,6 +4729,9 @@ components:
            processing results in the future.
         * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
         * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
+
+        If multiple processes with the same identifier exist, Clients SHOULD
+        inform the user that it's recommended to select a namespace.
     process_id:
       type: string
       description: |-
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4718,10 +4718,21 @@ components:
         * `null` (default): Checks both user-defined and pre-defined processes,
            but prefers user-defined processes if both are available.
            This allows users to replace missing pre-defined processes with
-           user-defined processes, e.g. processes from https://processes.openeo.org
-           that have a process graph included.
+           user-defined processes for portability, e.g. common processes from
+           [processes.openeo.org](https://processes.openeo.org) that have a process graph
+           included.
         * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
-        * `user`: Uses exclusively the user-defined processes listed at  listed as `GET /process_graphs`.  
+        * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
+
+        Note on collisions: The only way a collision between user-defined processes and
+        pre-defined processes can occur is when back-ends introduce a new process with an
+        identifier already used by users. User-defined processes with an identifier used by
+        a pre-defined process MUST be rejected. To ensure reproducibility, back-ends SHOULD
+        persist the namespace that is prioritized at the time the process is submitted.
+        If a collision occurs after a has back-end introduced a new pre-defined process
+        (i.e. a user-defined process is prioritized over a pre-defined process),
+        a log entry should be added so that the user is informed about the availability of the
+        new process.
     process_id:
       type: string
       description: |-
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4720,8 +4720,7 @@ components:
            This allows users to add missing pre-defined processes for portability,
            e.g. common processes from [processes.openeo.org](https://processes.openeo.org)
            that have a process graph included.
-           If a user-defined process is prioritized over a pre-defined process,
-           a log entry SHOULD be added.
+           It is RECOMMENDED to log the namespace selected by the back-end for debugging purposes.
         * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
         * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
 
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4723,9 +4723,10 @@ components:
            included.
            If a user-defined process is prioritized over a pre-defined process,
            a log entry SHOULD be added.
-           To ensure reproducibility, back-ends SHOULD persist the namespace that
-           is prioritized at the time the process is submitted so that adding or
-           removing processes doesn't change the processing results in the future.
+           To improve reproducibility, it is RECOMMENDED that back-ends persist
+           the namespace that would be selected at the time the process is
+           submitted so that adding or removing processes doesn't change the
+           processing results in the future.
         * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
         * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
     process_id:


openapi.yaml
```diff
@@ -4711,16 +4711,24 @@ components:
...
+           that have a process graph included.
+        * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
+        * `user`: Uses exclusively the user-defined processes listed at  listed as `GET /process_graphs`.  
     process_id:
       type: string
       description: |-
         Unique identifier for the process.
         
         MUST be unique across its namespace (e.g. pre-defined processes or user-defined processes).
+        Additionally, user-defined processes with an identifier equal to any pre-defined process MUST be rejected.
```
soxofaan, line 4731: Why is this necessary? It implies that namespaces are not independent, which breaks the point of namespaces.

m-mohr: Just to avoid confusion. Maybe users don't realize there's a process with the name and then with the default behavior in PGs it seems a bit less error-prone to just reject them. We could make this a recommendation only?!

soxofaan: It seems unnecessary to me because it's possible to refer to either one with the namespace property, so there is no unavoidable collision.
Also: even if you make it a recommendation in the API, a backend still has to make a hard choice what to do with it: accept  or reject (there is no channel for warnings or something like that)

soxofaan:
As discussed with @m-mohr in private: I'm not a big fan of re-introducing this behaviour of doing automagic name resolutions by default. The core problem I have is the practical disconnect between the place where you call the process (in a process graph) and where the decision lies between picking the PDP or UDP implementation (depending on whether the process is defined as UDP).  For example, when working in a language like Python, which  allows to overwrite predefined functions, it is still pretty clear you are working with a replaced version because the overwriting happens in the same file, or it is imported in the same file. With the proposal on how to do this in openEO, it is far from clear that a call to e.g. `normalized_difference` will be a PDP or UDP, because it depends on some user/time-specific state in another part of the API. When a user complains "my process graph doesn't work" and provides this process graph, it will be annoying to investigate/reproduce because each process could be overwritten implicitly.

Another related problem is that it allows users to "overload" processes, without overloading rules (e.g. parameters should be same, return type should be the same, ...), which could ruin longer term maintainability

The bottom line is probably that I prefer an easily determinable  namespace to avoid confusion and surprise.

That being said, I'm not against making the behaviour possible, I just don't like it as default. It could be an opt-in feature, e.g. with a special namespace "auto". At least it would be explicit then.


--

Also: you could also question the rule to prefer UDP over PDP in the  original use case that started this issue:  
> User could simply replace missing pre-defined processes (e.g. normalized_difference) with our "alternative" processing instructions in openeo-processes

why then not prefer a PDP (probably implemented more efficiently, but possibly missing) over UDP (a best effort fallback)?

m-mohr:
There are several things I'd like to enable here:
1. reproducibility - adding a new process shouldn't change the old jobs etc. To ensure this the following is added to the API:
  - UDP can't override existing PDP 
  - UDP are preferred over PDP (so that back-ends are not limited by existing PDP when adding processes)
2. portability - transferring processes between backends should be easy. Thus, users should be able to fill gaps in PDP by submitting the "replacement" processes whenever possible as UDP (this was forgotten in #309)
3. Avoid confusion on user-side - Accidentally adding a UDP with the name of a PDP can lead to confusing situations.

Getting back to @soxofaan's concerns:

> because it depends on some user/time-specific state in another part of the API. When a user complains "my process graph doesn't work" and provides this process graph, it will be annoying to investigate/reproduce because each process could be overwritten implicitly.

Could it help to specify that the back-ends can/should/must persist what namespace is correct at the time of submission?

> Another related problem is that it allows users to "overload" processes, without overloading rules

No, I want to avoid exactly that with: "Additionally, user-defined processes with an identifier equal to any pre-defined process MUST be rejected."

> That being said, I'm not against making the behaviour possible, I just don't like it as default.

But that's exactly what enables portability. Users will usually not care and wonder about issues then (I guess). I'm in favor of making it easy for users and if they want it explicit, it can be set.

> why then not prefer a PDP (probably implemented more efficiently, but possibly missing) over UDP (a best effort fallback)?

To not ruin reproducibility. If only a back-end adding a process can lead to conflicts, it has to prefer the process existing first (UDP). What I'd recommend it that back-ends throw a warning in the logs on conflicts and tell users there may be a faster alternative as PDP.

m-mohr:

m-mohr:
I added a bit more explanations in the last commit.

openapi.yaml
```diff
@@ -4711,16 +4711,35 @@ components:
...
+
+        Note on collisions: The only way a collision between user-defined processes and
+        pre-defined processes can occur is when back-ends introduce a new process with an
+        identifier already used by users. User-defined processes with an identifier used by
+        a pre-defined process MUST be rejected. To ensure reproducibility, back-ends SHOULD
+        persist the namespace that is prioritized at the time the process is submitted.
+        If a collision occurs after a has back-end introduced a new pre-defined process
+        (i.e. a user-defined process is prioritized over a pre-defined process),
+        a log entry should be added so that the user is informed about the availability of the
+        new process.
```
lforesta, line 4735: > a log entry should be added
Would this be available under /jobs ? Assuming th job works, I doubt most users will ever look in there, so they might not be aware that a new process exists.

soxofaan: I agree that a log entry is not a very strong signal

soxofaan:
> Could it help to specify that the back-ends can/should/must persist what namespace is correct at the time of submission?

with "is correct" I assume you mean "which namespace was picked"?

What would this also mean in practice? If you persist it, it must also be retrievable in some way, preferably through a standardized API I guess

openapi.yaml
```diff
@@ -4711,16 +4711,35 @@ components:
...
+           user-defined processes for portability, e.g. common processes from
+           [processes.openeo.org](https://processes.openeo.org) that have a process graph
+           included.
+        * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
+        * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
+
+        Note on collisions: The only way a collision between user-defined processes and
+        pre-defined processes can occur is when back-ends introduce a new process with an
+        identifier already used by users. User-defined processes with an identifier used by
+        a pre-defined process MUST be rejected. To ensure reproducibility, back-ends SHOULD
```
soxofaan, line 4730: I don't like this rule "collisions are ok if UDP was first". I don't think this will scale well if other namespaces are added. I would just don't care about "collisions" on this level because there is no collision: there are in different namespaces.

as noted by @m-mohr client could still raise a warning if they detect this kind of name collision when creating a UDP, which sound better because the API itesel has no "warning" channel

openapi.yaml
m-mohr: Yes, in the job or service logs. It will likely not work or give unexpected results, I guess. Otherwise the processes would be at least quite similar anyway (parameter names and schemas must match).

openapi.yaml
m-mohr: > I don't like this rule "collisions are ok if UDP was first". I don't think this will scale well if other namespaces are added. I would just don't care about "collisions" on this level because there is no collision: there are in different namespaces.

This is meant to only apply between PDP and UDP. Other namespace are not involved.

> as noted by @m-mohr : the client could still raise a warning if they detect this kind of name collision when creating a UDP, which sound better because the API itself has no "warning" channel

It has the log files endpoints though.

m-mohr:
> with "is correct" I assume you mean "which namespace was picked"?

Yes, the one picked by the back-end.

> What would this also mean in practice? If you persist it, it must also be retrievable in some way, preferably through a standardized API I guess

Yes, request GET /jobs/jobid.

openapi.yaml
soxofaan: > It has the log files endpoints though

also for UDP creation?

m-mohr: No, just for jobs/services indeed.

lforesta: I like the idea of having the client raising a warning, since users interface with that rather than the back-end

soxofaan: > This is meant to only apply between PDP and UDP. Other namespace are not involved.

indeed, and that's why it does seem to be so arbitrary to me. Or you forbid all possible collisions, or you allow them everywhere, something in between feels hackish and ad-hoc.

m-mohr: Okay, I'm going to replace the UDP rejection rule with the following:

> Clients SHOULD warn the user if a user-defined process is added with an identifier that exists in the pre-defined processes.

soxofaan:
>  >    What would this also mean in practice? If you persist it, it must also be retrievable in some way, preferably through a standardized API I guess
>
> Yes, request GET /jobs/jobid.

I think this will incur a much larger change for backend to manage and implement than the backward incompatible change that was introduced by #309 in the first place

openapi.yaml
m-mohr: See https://github.com/Open-EO/openeo-api/pull/313/commits/59703ae6d82faebd44b8951a735fac64bdf1a4b8

m-mohr:
Okay, next attempt. Please review it again.

lforesta:
> > > What would this also mean in practice? If you persist it, it must also be retrievable in some way, preferably through a standardized API I guess
> > 
> > 
> > Yes, request GET /jobs/jobid.
> 
> I think this will incur a much larger change for backend to manage and implement than the backward incompatible change that was introduced by #309 in the first place

Also, how would this work in practice? Each time there is a PDP/UDP conflict, the back-end should find the oldest job from that user containing the conflicting UDP/PDP and extract the namespace from that?

openapi.yaml
m-mohr: Has been replaced by a "clients should warn".

openapi.yaml
m-mohr: Now clients should warn, but back-ends can should still also warn, I think.

m-mohr:
> Also, how would this work in practice? Each time there is a PDP/UDP conflict, the back-end should find the oldest job from that user containing the conflicting UDP/PDP and extract the namespace from that?

I'd just set the namespaces in the PG on submission.

openapi.yaml
```diff
@@ -4711,16 +4711,31 @@ components:
...
+        * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
+        * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
     process_id:
       type: string
       description: |-
         Unique identifier for the process.
         
         MUST be unique across its namespace (e.g. pre-defined processes or user-defined processes).
+        Clients SHOULD warn the user if a user-defined process is added with an identifier that
+        exists in the pre-defined processes.
```
soxofaan, line 4738: suggestion to make it bit simpler:

> Clients SHOULD warn the user if a user-defined process is added with the same identifier as a pre-defined process.

lforesta:
> I'd just set the namespaces in the PG on submission. You're validating it anyway...

I thought we were discussing mostly the option where the namespace is 'null' and the back-end needs to select the correct process from a possible UDP/PDP conflict, no? If the namespace is set to a specific value, there is no ambiguity.

openapi.yaml
m-mohr: Done, thanks.

openapi.yaml
```diff
@@ -4711,16 +4711,31 @@ components:
...
+           but prefers user-defined processes if both are available.
+           This allows users to replace missing pre-defined processes with
+           user-defined processes for portability, e.g. common processes from
+           [processes.openeo.org](https://processes.openeo.org) that have a process graph
+           included.
+           If a user-defined process is prioritized over a pre-defined process,
+           a log entry SHOULD be added.
+           To ensure reproducibility, back-ends SHOULD persist the namespace that
+           is prioritized at the time the process is submitted so that adding or
+           removing processes doesn't change the processing results in the future.
```
soxofaan, line 4728: > To ensure reproducibility, backend SHOULD persist ...so that .. doesn't chang the processing results ...

This  reads a bit vague to me, but it still sounds that there is quite some responsibility for a backend regarding reproducibility features. Like an old process graph should still run in some kind of snapshotted UDP/PDP context?

m-mohr:
> I thought we were discussing mostly the option where the namespace is 'null' and the back-end needs to select the correct process from a possible UDP/PDP conflict, no? If the namespace is set to a specific value, there is no ambiguity.

I meant: set by the back-end, not by the user. I.e. user sends with namespace null, back-end persists on submission to user or back-end so that on future changes nothing breaks.

openapi.yaml
soxofaan: also if you speak about persisting: persist based on what handle or identifier? for example: jobs and services have an  identifier, but for synchronous `/result` call there is no identifier .

m-mohr: Yes, this is more a recommendation to the back-end though. At least I'd expect that at least PDP give the same results after a year and also UDP unless explicitly altered by a user. If a PDP changes, you are lost as a user anyway, with or without namespaces.

m-mohr: Maybe it should read:

```> To *improve* reproducibility, it is RECOMMENDED that back-ends persist the namespace that
           is prioritized (selected?) at the time the process is submitted so that adding or
           removing processes doesn't change the processing results in the future.

Comment (m-mohr) [reply openapi.yaml]: Made some changes in the last commit to not put too much pressure regarding reproducibility.

Comment (soxofaan) [reply openapi.yaml]: Wouldn't it just be easier to say that if reproducibility is important for the user it is recommended to specify the namespace explicitly instead of depending on run-time name resolution details?

Comment (m-mohr) [reply openapi.yaml]: User usually don't read the API spec and as a user I'd expect something like that from the back-end tbh.

Comment (soxofaan) [reply openapi.yaml]: Also, another sidenote: I don't think that UDP are the best tool for users that are very wary about reproducibility: you only have a single user-level global UDP namespace. In modern development it's typically recommended to work in per-project sandboxes (e.g. virtual envs in python, etc). 
I'm not saying we should allow users to define multiple namespaces (yet :smile:  ), but if I would need a highly reproducible and stable workflow with openEO, I think I would prefer to just embed everything in a large PG instead of managing and calling UDPs.

Comment (lforesta) [reply openapi.yaml]: Personally, I'm not a huge fan of this time-based namespace resolution and am even less of a fan that it is a recommendation only. In practice, a user then can't be sure what's happening or if the same approach is used on two different back-ends.
As a user I'd prefer to having to add/update a namespace in my PG and hence ensure that the back-end will run the job I want it to run, rather than leaving the namespace blank and crossing my fingers..

Comment (m-mohr) [reply openapi.yaml]: How would you update that? Except for the Web Editor there's no client AFAIK that allows to generate code from a PG, so a user would need to edit the PG, which is not something we want to reply on.

Comment (m-mohr) [reply openapi.yaml]: > I'm not saying we should allow users to define multiple namespaces (yet 😄 ), but if I would need a highly reproducible and stable workflow with openEO, I think I would prefer to just embed everything in a large PG instead of managing and calling UDPs.

Indeed, it would probably be a good idea if clients could just "resolve" UDP and create on big PG to be submitted. Then you still rely on PDP to not change, but maybe versioned PDP become a thing at some point 😄

Comment (soxofaan) [reply openapi.yaml]: with the Python client (and I guess the R client too) you express the PG in Python code, so it's trivial to edit/update that.

Also at VITO we have a user that doesn't want to depend too much on the Python client because it's not stable enough: he just does HTTP calls directly (from python), and manages a JSON file somewhere

Comment (soxofaan) [reply openapi.yaml]: Also: if you're using openEO in a serious way, at some point you probably are going to want to download some kind of representation of the PG in some way (JSON probably), for example to put it in version control or attach it to a report. So I don't think we should shield users too much from JSON. (If it would be XML, that would be another story :smile: )

Comment (m-mohr) [reply openapi.yaml]: > with the Python client (and I guess the R client too) you express the PG in Python code, so it's trivial to edit/update that.

Short term, yes, usually. Long term, probably less likely. And very unlikely for shared processes, which is the main reason to allow the `null` namespace.

> Also at VITO we have a user that doesn't want to depend too much on the Python client because it's not stable enough: he just does HTTP calls directly (from python), and manages a JSON file somewhere

Interesting, but I doubt that's not the normal use case ;-)

Comment (lforesta) [reply openapi.yaml]: imho the clients must set the namespace explicitly. 
When the user connects to a back-end, the clients query /processes and /process_graphs to get the PDPs and UDPs available at the back-end for that user. If the user selects a process for which there is a clash, the user should be asked to flag which process to use. The clients then sets the namespace accordingly before submitting the PG to the back-end.

Comment (m-mohr) [reply openapi.yaml]: I'm not against setting it explicitly when there's a clash, but unless there is a clash, the namespace should not be set.
The namespace should not be set by default so that sharing and importing PGs works better.

Comment (soxofaan) [reply openapi.yaml]: > The clients then sets the namespace accordingly before  ...

that would be the most user friendly, but you still need to specify back-end side behavior because there might be users that don't use (official) clients.

Comment (m-mohr) [reply openapi.yaml]: So should I add:

> Clients SHOULD ask the user to selects the namespace, if multiple processes with the same name exist.

Comment (soxofaan) [reply openapi.yaml]: > And very unlikely for shared processes (e.g. Hub),

To me that doesn't sound like a big problem. When I copy-paste code snippets from, say stackoverflow or some blog, I don't expect it to work automagically. There is always some  adaptation to do (changing variable names, fixing imports, ...).  We have this problem anyway with collection ids for example.

Comment (m-mohr) [reply openapi.yaml]: But that's what we made UDP for, that they can easily be shared. Like a library. You usually don't change libraries. I'm not really thinking about code snippets here, but libraries. And Collection IDs in UDPs should be parameters.

Comment (soxofaan) [reply openapi.yaml]: > Clients SHOULD ask the user to selects

that sound quite oriented towards interactive clients like the web editor. That doesn't work nice for clients that are expected to work in some kind of headless/batch mode (e.g. python scripts)

Comment (m-mohr) [reply openapi.yaml]: > that sound quite oriented towards interactive clients like the web editor. That doesn't work nice for clients that are expected to work in some kind of headless/batch mode (e.g. python scripts)

Why is that a problem?

> If you have ndvi as UDP and PDP and you call let's say builder$ndvi(data) then it prints/throws (whatever you like) a warning and could recommends you to set the namespace to null, user or backend. Should work in R, I guess also at least in Python Notebooks?

Comment (soxofaan) [reply openapi.yaml]: UDPs are a bit like libraries, but a lot more individual and granualar than how libraries work in other platforms.
I think a better equivalent of "libararies" in openeEO would be a using a URL as namespace, pointing to a repository of reusable processes.

(which is an approach we are thinking about at VITO for a use case we started working on)

Comment (m-mohr) [reply openapi.yaml]: Okay, maybe it's more like a library of functions, but you also don't go ahead and change functions in a library. You just use them according to the contract/interface...

Comment (soxofaan) [reply openapi.yaml]: "user-defined" does sounds like I have permissions to do what I want

Comment (m-mohr) [reply openapi.yaml]: Sure, you can submit whatever you like. You can also fork libraries on GitHub, but that's not how they are usually used. 

If you go to the Hub and there's a process you'd like to use, it's just a process that you can submit as a "user-defined" process to your back-end.

Comment (soxofaan) [reply openapi.yaml]: thinking a bit out of the box here: could we add some kind of "name resolution strategy" property somewhere (e.g. as parameter in job/service/result request, or as field in `process` JSON body), so that user/client can pick a strategy that is best for the usecase or context?

Comment (m-mohr) [reply openapi.yaml]: > yes, you can use python client in interactive contexts. But my point is that if I have a script which I run daily with some cron system, I don't want it to suddenly crash/hang one day because the backend introduced a "collision"

Okay, maybe it's better to say:
> If multiple processes with the same identifier exist, Clients SHOULD inform the user that it's recommended to select a namespace.

Then it shouldn't crash

Comment (m-mohr) [reply openapi.yaml]: > thinking a bit out of the box here: could we add some kind of "name resolution strategy" property somewhere (e.g. as parameter in job/service/result request, or as field in `process` JSON body), so that user/client can pick a strategy that is best for the usecase or context?

So that it would allow to let's say prefer PDP over UDP?

Comment (soxofaan) [reply openapi.yaml]: indeed, 
and in the longer term if more namespaces are defined (officially or customly by a backend): in which order to scan the namespaces

Comment (soxofaan) [reply openapi.yaml]: for example: it's a list like `["user", "backend"]` to say to prefer UDP over backend. and if the list is empty or null: you error on any collision

Comment (m-mohr) [reply openapi.yaml]: I thought you were aiming to simplify and remove the default behavior, but that sounds even more complex and time-based ;-)  There are only really UDP and PDP in the resolution strategy, all others are likely having namespaces (usually URLs) anyway, I guess.

Comment (soxofaan) [reply openapi.yaml]: and it would only apply for processes where namespace is not set

Comment (m-mohr) [reply openapi.yaml]: Okay, hmm... I don't really see the advantage, I must be missing something.
There's only really three cases: prefer UDP, prefer PDP and error
Currently we do prefer UDP and I thought you were not liking error. Prefer PDP currently is only possible by setting the namespace for a new process or removing the UDP. Does that really need a resolution strategy parameter?

Comment (lforesta) [reply openapi.yaml]: there is still the edge case where a user has multiple UDPs in his/her PG, some of which now clash with PDPs, and they want only a subset of them to be used as UDP. In this case they must specify the namespace explicitly.

Comment (soxofaan) [reply openapi.yaml]: > all others are likely having namespaces

Indeed that's true currently. However, it could be extended to for example `"name_resolution_order"=["http://url/to/repo", "backend"]`, which would mean that processes without an explicit namespace are first looked up at that url.

> but that sounds even more complex and time-based

true, it indeed smells like feature creep. Just trying to find a way out of this ~~conflict~~ issue :smile:

Comment (m-mohr) [reply openapi.yaml]: I guess we can't solve all edge cases, but at least they can specify the namespace per process then.

Comment (soxofaan) [reply openapi.yaml]: > Currently we do prefer UDP and I thought you were not liking error by default

I don't like automagic name resolution by default. And I don't like the introduction of "conflict" as a thing to worry about. 

The current behavior per API without PR 313 is: use "backend" namespace by default if namespace is not set, and if it is set: use that. There is not concept of conflict whatsoever. And when the process can not be found: you have an error. Plain and simple

Comment (soxofaan) [reply openapi.yaml]: This PR 313 introduces the concepts "name resolution" and "name conflict" for reasons and use cases I'm not that famliar with, but for which my simple solution would be "just specify the namespace explicitly".

Comment (soxofaan) [reply openapi.yaml]: some reasons I don't like (automatic) name resolution:
- it is dependent on timing, user specific state (their set of UDPs), which makes thing harder to investigate and reason about
- we don't care about overloading rules or preserving the signature (input and output types) of a function, which should be part I think of a dynamic resolution system

Comment (m-mohr) [reply openapi.yaml]: > issue/conflict: how does the back-end resolve which process to use when `namespace` is null and there are one UDP and one PDP with the same id?

Currently it says UDP over PDP as it is most flexible: Users have two of options to avoid that: delete UDP or set namespace.

> Possible solution from @soxofaan : specify resolution_order as a parameter. If missing, return conflict errors

What is the advantage and what should be the default?

> This approach is not bad because it makes it explicit for back-ends which process to use (as oppposed to finding out which was set first between UDP and PDP)

What you describe here is not the behavior proposed in this PR. The PR doesn't say you need to find out what was first, it just says prefer UDP over PDP. That seems exactly as explicit?!

> I don't like automagic name resolution by default. And I don't like the introduction of "conflict" as a thing to worry about.

> The current behavior per API without PR 313 is: use "backend" namespace by default if namespace is not set, and if it is set: use that. There is not concept of conflict whatsoever. And when the process can not be found: you have an error. Plain and simple

Yes, but we forgot a use-case in the PR unfortunately. We introduced the process graphs to openeo-processes so that they can be used instead of missing pre-defined processes at back-ends. And I also didn't really took sharing into account there. So I'm now just trying to get (partly) back to RC2. There were means to avoid collisions in RC2 (reject UDP, prefer UDP oder (later introduced) PDP). But I removed that as you requested it, but now you don't want the conflicts, too. That's a bit confusing ;-)

**Comment by m-mohr** [comment]
I think I'm going to remove the recommendation regarding reproducibility. In the long run, there's more to do if you really want reproducibility. In this case you'd probably need to set a flag for the batch job that persists more things (UDP, namespaces, external processes, maybe even the version of the PDP etc.) That could be a new extension, maybe based on the previous work from @bgoesswe .

With that, we may have considered points from both sides. It doesn't seem there's a solution that makes everybody happy so that will probably be the compromise that is getting merged in the next hour.

**Comment by soxofaan** [comment]
> ... remove the recommendation regarding reproducibility. In the long run, there's more to do if you really want reproducibility. ...

I agree, that is not something we will be able to address properly just before the deadline

File: CHANGELOG.md [hunk 6]
```diff
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /.well-known/openeo` and `GET /`: `production` fields default to `false` instead of `true`.
 - `GET /collections` and `GET /collections/{collection_id}`: Additional dimensions in `cube:properties` can only be of type `other`.
 - Allow all STAC versions that are compatible to STAC 0.9.0.
-- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The old behavior to prioritize user-defined over pre-defined processes has been removed. [#305](https://github.com/Open-EO/openeo-api/issues/305)
+- Process graph nodes have an additional field `namespace` to distinguish pre-defined and user-defined processes. The old behavior to prioritize user-defined over pre-defined processes is still the default. [#305](https://github.com/Open-EO/openeo-api/issues/305)
```
soxofaan, line 25: > The old behavior to prioritize user-defined over pre-defined processes is still the default.

->
 When no namespace is given explicitly: user-defined processes have priority over predefined process (the old behavior)

openapi.yaml
```diff
@@ -4711,16 +4711,32 @@ components:
...
+        * `backend`: Uses exclusively the pre-defined processes listed at `GET /processes`.
+        * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
+
+        If multiple processes with the same identifier exist, Clients SHOULD
+        inform the user that it's recommended to select a namespace.
     process_id:
       type: string
       description: |-
-        Unique identifier for the process.
+        An identifier for the process.
```
soxofaan, line 4733: > An

Maybe even better `The`, because `An` implies there are multiple identifiers for the same process (unless that is what you mean).

openapi.yaml
```diff
@@ -4711,16 +4711,32 @@ components:
...
         
-        MUST be unique across its namespace (e.g. pre-defined processes or user-defined processes).
+        The identifier MUST be unique across its namespace (e.g. pre-defined
+        processes or user-defined processes).
+
+        Clients SHOULD warn the user if a user-defined process is added with the 
+        same identifier as one of the pre-defined process.
```
soxofaan, line 4739: This note about "should warn" is under the `namespace` field when describe process graphs right?  Shouldn't it be in the docs of creating a UDF?

openapi.yaml
```diff
@@ -4711,16 +4711,32 @@ components:
...
+        The following options are pre-defined by the openEO API, but additional
+        namespaces may be introduced by back-ends or in a future version of the API.
         
-        By default, the namespace is `null`, which includes all pre-defined processes listed at `GET /processes`.
-        The namespace `user` refers to all user-defined processes listed as `GET /process_graphs`.
-        Additional namespaces may be introduced by back-ends or in a future version of the API.
+        * `null` (default): Checks both user-defined and pre-defined processes,
+           but prefers user-defined processes if both are available.
+           This allows users to add missing pre-defined processes for portability,
+           e.g. common processes from [processes.openeo.org](https://processes.openeo.org)
```
soxofaan, line 4721: Is that URL https://processes.openeo.org stable and future-proof enough to be included in the API spec? Otherwise the `e.g. ...` can be dropped without much loss of information I think.

openapi.yaml
m-mohr: No, it's defined under the process id field.

openapi.yaml
m-mohr: The URL is already used in several places throughout the API. It should be stable enough, except that it doesn't contain a reference to a explicit process version.

openapi.yaml
m-mohr: Yes, will fix.

openapi.yaml
```diff
@@ -4711,16 +4711,32 @@ components:
...
+           that have a process graph included.
+           If a user-defined process is prioritized over a pre-defined process,
+           a log entry SHOULD be added.
```
soxofaan, line 4724: about adding log entries: wouldn't it be better to *recommend* to always log the resolution (so not only when UDP is picked). Because if user finds no log: it could mean two things:
- UDP was picked but backend did not implement the logging (it's a SHOULD)
- PDP was picked

soxofaan:
(sorry for the last minute review spam)

CHANGELOG.md
m-mohr: I think I'm going for: The default behavior still prioritizes user-defined over pre-defined processes.

openapi.yaml
soxofaan: I mean: the warning should happen on `PUT /process_graph/{}`  , not when evaluating a process graph (for batch jobs/services/synchronous). So I would put the `Clients SHOULD warn ..` closer to the `PUT /process_graph/{}` docs

CHANGELOG.md
soxofaan: The point of my suggestion was mainly to add `When no namespace is given explicitly` somewhere in there. I know it is obvious for us, actors in this PR, but maybe less obvious for others.

openapi.yaml
m-mohr: recommend vs should: According to the RFC, it has the same meaning.

I have no strong reason against adding/changing that. How would you express that?

openapi.yaml
m-mohr: And yes, I'll add it there in addition.

openapi.yaml
soxofaan: ok I didn't realize that, to me "recommend" sound bit less forcing than "should".

m-mohr:
> (sorry for the last minute review spam)

Thanks, I appreciate the last minute review. It brought me to read other outdated party again and I could update them.

openapi.yaml
soxofaan: Also the sentence should state that it's  back-end-side logging

m-mohr:
I removed a bit of namespace description here: https://github.com/Open-EO/openeo-api/pull/313/commits/84f5ce49784edd23c2b86e0731c346a2f21a0084#diff-fe030a7c1568b3decf599edf399be7f4R276

I think I don't want to repeat the whole namespace description there.

openapi.yaml
soxofaan: suggestion:

> The back-end SHOULD add a log entry about whether the user-defined process or the predefined-process was picked to simplify post-mortem analysis.

minor note: I would tend to use MAY instead of SHOULD however, because the logging is not a first class feature of the API at this point, I would say

m-mohr: > To me, "recommend" sound bit less forcing than "should".

Yeah, same for me.

> Also the sentence should state that it's back-end-side logging

Hmm... "a log entry SHOULD be added to corresponding back-end resource"?

soxofaan: > The backend SHOULD add a log entry to the corresponding resource about whether the user-defined process or pre-defined process was picked, for debugging reasons.

m-mohr: And for general logging, maybe: 

> It is RECOMMENDED to log the selected namespace for each process to the corresponding back-end resource.

m-mohr: Oh, haven't seen yours early enough. Which one do you prefer? Mine seems to be a bit shorter...

CHANGELOG.md
m-mohr: Hmm... is "namespace = null" a "given namespace"?

soxofaan: I think for a CHANGELOG entry it shouldn't be very pedantic about that

soxofaan: > When the user didn't specify a namespace ...

m-mohr: That doesn't quite express what I intend to say there. I want to express that this is more an addition, not a breaking change (the default behavior is still the same). That is somewhat lost...

Issue #315, opened 2020-07-14
batch job logs when job is not finished or even started


soxofaan:
`GET /jobs/{job_id}/logs`

It's not specified explicitly how back-end should behave when a batch job is still "created"/"queued".
In VITO backend we currently throw `JobNotStarted` error, because that seemed fitting, but (D28) validator does not like that.
Should we just return empty list of log entries?

m-mohr:
Logs can be used at all stages of a job, but I see that this part can be a bit misleading:

> If requested consecutively while a job is running, it is RECOMMENDED that clients use the offset parameter to get only the entries they have not received yet.

I'll try to improve and issue a PR.

m-mohr:
see PR #316

Issue #314, opened 2020-07-13
Authorization of /validation


soxofaan:
https://github.com/Open-EO/openeo-api/blob/12471721d9c74262ae962f7ad6c746ac77aad382/openapi.yaml#L2195-L2229

is it intended that `/validation`  can also be requested without authorization?

m-mohr:
Yes, we thought it's helpful if users could validate and check whether the back-end is capable of running their stuff without registering and all the hassle.

soxofaan:
ok make sense.

One minor thing is that `/validation` is currently only under category `User-Defined Processes` while from your reply I understand it is about validation of complete process graphs (as well). They are very much alike, but not completely: e.g. in a UDP it is fine to have unresolved parameters, while in a complete process graph you don't want that, which could be a relevant validation aspect.

m-mohr:
Hmm, I now realized it's not clear from the description what to do with unresolved parameters. I'd argue that unresolved parameters shouldn't lead to an error (and thus it's in this category). Setting the parameters etc seems to me more fitting into submitting the batch job/service/... 

We should clarify that. If we wan't to check for unresolved parameters, I'd guess the best idea is adding a parameter to the request?

soxofaan:
Well my question is also: if it's only about UDPs, is `/validation` the best url for that, as all the other UDP urls start with `/process_graphs`?

m-mohr:
There are some things to consider:
1. The processes you send to batch jobs/services/... are still UDP techically. There's very little difference schema-wise (just in metadata).
2. /process_graphs/validation doesn't work due to the conflict with the id parameter
3. I'm not very happy with the legacy /process_graphs anymore anyway. It seems you are neither: #310
4. If we want to add validation including unresolved parameters, a simple backward compatible path parameter could be added.

So I think it's still reasonable to just call it /validation.

soxofaan:
ok, so to conclude: the `/validation` endpoint is not only about UDPs in the sense of the named "user defined processes" with a life cycle (storage and retrieval), but also one-off "anonymous" process graphs (e.g. for batch jobs).
Shouldn't `/validation` then also be tagged under the generic tag "Data processing"?

I'm totally fine with adding parameters to `/validation` to fine-tune the validation mode/strategy. In VITO backend we haven't implemented it so far, so we don't have practical insights here.

m-mohr:
Made the changes in PR #316

PR #316, opened 2020-07-14
Clarifications for #314, #315

m-mohr:
see #314 and #315

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1878,26 +1878,24 @@ paths:
     post:
       summary: Validate a user-defined process (graph)
       operationId: validate-custom-process
-      description: >-
+      description: |-
         Validates a user-defined process without executing it. A user-defined process is
         considered valid unless the `errors` array in the response contains at
         least one error.
 
-
         Checks whether the user-defined process is schematically correct and the
         processes are supported by the back-end. It MUST also checks the
-        parameter values against the schema, but checking whether the values are
+        arguments against the schema, but checking whether the arguments are
         adequate in the context of data is OPTIONAL. For example, a non-existing
         band name may get rejected only by a few back-ends.
 
+        The validation MUST NOT throw an error for unresolvable process parameters.
 
         Errors that usually occur during processing MAY NOT get reported, e.g.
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -1878,33 +1878,32 @@ paths:
...
       operationId: validate-custom-process
-      description: >-
+      description: |-
         Validates a user-defined process without executing it. A user-defined process is
         considered valid unless the `errors` array in the response contains at
         least one error.
 
-
         Checks whether the user-defined process is schematically correct and the
         processes are supported by the back-end. It MUST also checks the
```
soxofaan, line 1887: checks -> check

openapi.yaml
```diff
@@ -1878,33 +1878,32 @@ paths:
...
       summary: Validate a user-defined process (graph)
       operationId: validate-custom-process
-      description: >-
+      description: |-
         Validates a user-defined process without executing it. A user-defined process is
         considered valid unless the `errors` array in the response contains at
         least one error.
 
-
         Checks whether the user-defined process is schematically correct and the
```
soxofaan, line 1886: user-defined process -> process graph ?

m-mohr: user-defined process includes the process graph, but validators may also validate the metadata.

openapi.yaml
m-mohr: Fixed

openapi.yaml
m-mohr: Clarified it a bit more.

soxofaan: To me,  UDP means a process graph with a name (because the `process_graph_id` it is required in the UDP life cycle), while `/validate` also covers "anonymous" process graphs (the ones you pass as batch jobs or to `/execute`).

maybe write "process graph or user-defined process" ?

m-mohr: See https://github.com/Open-EO/openeo-api/pull/316/commits/8d7aaf82ac04b3fefbd766b0f3ea51f8af58a348 - I think my changes and your comment overlapped.

PR #319, opened 2020-07-14
Fine-tune secondary service examples

m-mohr:
Looks a bit long in the OpenAPI file, but is pretty sure once rendered.


m-mohr:
Oh, really? Seems I take them for granted because I worked with them in the studies. What example would you have thought of?

soxofaan:
no I mean I'm a relative newcomer to the world of GIS in general, I only started at VITO last year :)

PR #321, opened 2020-07-15
Parameters for File Formats / Service Types

m-mohr:
The Parameters for File Formats and Service Types had a JSON Schema based definition, but it was a (too) limited subset. It's now replaced with JSON Schema itself. It's used in other places anyway so code can better be re-used. For back-ends this is not a breaking change, it's fully compatible. Clients may face additional/different schemas (especially an array for type), but it's not a bigger issue, I think.

Additionally, I added descriptions for the JSON Schema properties with links to the actual JSON Schema spec.

This also fixes an invalid enum-example in the /service_types response.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2244,7 +2244,7 @@ paths:
                       description: Map of supported configuration settings made available to the creator of the service.
                       type: object
                       additionalProperties:
-                        $ref: '#/components/schemas/argument'
+                        $ref: '#/components/schemas/resource_parameter'
                     process_parameters:
                       title: Process Parameters
                       description: List of parameters made available to user-defined processes.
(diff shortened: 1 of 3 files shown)


m-mohr:
Please review this regardless of the failing CI, that's a bug in the linter/CI, not in the spec itself.

openapi.yaml
```diff
@@ -5114,67 +5116,43 @@ components:
...
+      description: |-
+        Describes a parameter for various resources (e.g. file formats, service types).
+
+        The parameters are specified according to the [JSON Schema draft-07](http://json-schema.org/) specification.
+        See the chapter ['Schemas' in 'Defining Processes'](#section/Processes/Defining-Processes) for more information.
+
+        It is recommended to refrain from using the following more complex JSON Schema keywords:
+        `if`, `then`, `else`, `readOnly`, `writeOnly`, `dependencies`, `minProperties`, `maxProperties`, `patternProperties`, `multipleOf`.
+
+        JSON Schemas should always be dereferenced (i.e. all `$refs` should be resolved). This allows clients to consume the schemas much better.
```
soxofaan, line 5132: shouldn't the "should" and "recommended" (a bit higher) be in upper case?

m-mohr: Yes, indeed. Fixed it. Also made commit bc15fe17 to do some similar changes.

m-mohr:

Issue #318, opened 2020-07-14
No requirements for GET /service_types


soxofaan:
The service entries in `GET /service_types` currently have no required fields: each of `configuration`, `process_parameters` and `links` is allowed to be missing.

This caused the validator to accept the 0.4-style version of this endpoint in the 1.0-version of the VITO backend, e.g. currently at http://openeo.vgt.vito.be/openeo/1.0/service_types: 
```
{
    "WMTS": {
        "attributes": [],
        "parameters": {
            "colormap": {
                "default": "YlGn",
                "description": "The colormap to use.",
                "type": "a colormap to apply to single band layers"
            },
            "version": {
                "default": "1.0.0",
                "description": "The WMTS version to use.",
                "enum": [  "1.0.0"  ],
                "type": "string"
            }
        }
    }
}
```

Note these errors (that the validator didn't catch):
- 0.4-style fields `attributes` and `parameters`
- invalid `"type": "a colormap to apply to single band layers"`

Can this be avoided in some way?
For example require that `process_parameters` is present with at least and empty array?

soxofaan:
cc @bgoesswe

m-mohr:
Due to the "extensibility" of JSON, it's hard to avoid.

I'd actually like to require configuration and process_parameters as it hopefully makes the back-end think what it needs to set ;-) Not sure whether we should introduce the breaking change in the release week though.

It doesn't catch the invalid type as it's in an "undefined" structure, but it would be caught if it's in process_parameters.

m-mohr:
What I guess would be useful to have in the validator, that it reports "undefined" properties as warnings or so. @bgoesswe

soxofaan:
> Not sure whether we should introduce the breaking change in the release week though with just the reason to help the validator.

according to the openeo hub the support for secondary services is pretty thin (and so I assume practically unused), so I don't think a lot would break actually. But indeed it's already late to do these kind of things

m-mohr:
Indeed, just VITO and GEE. I'm fine with it and if VITO is fine with it, too.

bgoesswe:
> What I guess would be useful to have in the validator, that it reports "undefined" properties as warnings or so. @bgoesswe

Yep that seems like a good idea, I added a [new issue](https://github.com/Open-EO/openeo-backend-validator/issues/39) at the validator for that.

soxofaan:
>  if VITO is fine with it

[...]

PR #322, opened 2020-07-15
Require Service Type Config

m-mohr:
see #318

I just required configuration, process_parameters just applies to a subset of services, I think. If we's require process_parameters, it would also be good to require parameters in file_formats, I guess.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /collections/{collection_id}`: Added optional `assets` property for collection-level assets. This may link to visualizations for example. [#211](https://github.com/Open-EO/openeo-api/issues/211)
 - `GET /collections`, `GET /jobs`, `GET /process_graphs`, `GET /Services`: Allow all non-scalar properties to be part of the response although strongly discouraged.
 
-### Changes
+### Changed
 - `GET /credentials/oidc`: field `scopes` is not required anymore, but when specified, it should contain the `openid` scope. [#288](https://github.com/Open-EO/openeo-api/pull/288)
 - `GET /.well-known/openeo` and `GET /`: `production` fields default to `false` instead of `true`.
 - `GET /jobs/{job_id}/logs` and `GET /services/{service_id}/logs`: `path` property is not required any longer. [#320](https://github.com/Open-EO/openeo-api/issues/320)
+- `GET /service_types`: `configuration` is now required for each service. [#318](https://github.com/Open-EO/openeo-api/issues/318)
 - `GET /collections` and `GET /collections/{collection_id}`:
     - Additional dimensions in `cube:properties` can only be of type `other`.
     - The extents `interval` and `bbox` can have multiple entries.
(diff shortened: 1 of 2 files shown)


soxofaan:
If a field is required, it still can be an empty object (for `configuration`) or empty array (for `process_parameters`), right?

I wouldn't mind also requiring `process_parameters` as it makes it more explicit that a service does not expect parameters for example.

m-mohr:
Yes, both can be empty.

> I wouldn't mind also requiring process_parameters as it makes it more explicit that a service does not expect parameters for example.

I agree, but as this also applies to parameters in file_formats, we should also change it there. Not sure whether that's wanted or not.

soxofaan:
> also applies to parameters in file_formats, we should also change it there. Not sure whether that's wanted or not.

hmm indeed, making it mandatory there is less favourable.  Is it tedious (openapi-yaml-wise) to decouple service parameters and file format parameters?

m-mohr:
They are decoupled, but I think both are very similar and thus I think if it's required in /service_types, it should also be required in file_formats. I just don't see a good (spec-related) reason to handle them differently.

soxofaan:
Schematically they are similar, but conceptually quite different in the context of openEO I would think:
- process_parameters in services are *process parameters* which is a well established concept in openEO and has important connotations, for example a service without parameters (e.g. to change bbox/date) lacks interactivity/dynamism, which seems useless
- the parameters in a file format are more obscure and probably just for power users that want to override defaults.

That said, making parameters required for file formats as well is not that big of an issue I think, it's a very simple fix for backend to implement (just add `"parameters":[]` for a parameter-less format). And it forces backends to think about providing parameters for their formats in the first place, which is a good thing.

m-mohr:
Agreed. I added the additional requirements.

m-mohr:

CHANGELOG.md
```diff
@@ -17,10 +17,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /collections/{collection_id}`: Added optional `assets` property for collection-level assets. This may link to visualizations for example. [#211](https://github.com/Open-EO/openeo-api/issues/211)
 - `GET /collections`, `GET /jobs`, `GET /process_graphs`, `GET /Services`: Allow all non-scalar properties to be part of the response although strongly discouraged.
 
-### Changes
+### Changed
 - `GET /credentials/oidc`: field `scopes` is not required anymore, but when specified, it should contain the `openid` scope. [#288](https://github.com/Open-EO/openeo-api/pull/288)
 - `GET /.well-known/openeo` and `GET /`: `production` fields default to `false` instead of `true`.
 - `GET /jobs/{job_id}/logs` and `GET /services/{service_id}/logs`: `path` property is not required any longer. [#320](https://github.com/Open-EO/openeo-api/issues/320)
+- `GET /file_formats`: `parameters` is now required for each file format. [#318](https://github.com/Open-EO/openeo-api/issues/318)
+- `GET /service_types`: `configuration` and `process_parameters` are now required for each service. [#318](https://github.com/Open-EO/openeo-api/issues/318)
```
soxofaan, line 25: > ... now required for ...

maybe add `(but can be empty object/list)` to reduce friction for implementers ?

m-mohr: Not sure the CHANGELOG is the right place for it. Except for removals, we never gave any further advice, I think.

With the background that we changed it so that people actually think about what to do with it, I think I'm against specifying this here.

PR #345, opened 2020-11-19
Fix security-related CORS issues regarding to credentials #41

### LINKED ISSUE #41: Authentication: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*'
Seems we have a severe problem with Authentication:
When sending Basic Auth credentials (probably also for Bearer token), the servers are not allowed to set the Access-Control-Allow-Origin to `*` (wildcard).

Not sure yet how this can be solved appropriately, maybe we need to have an alternative way for logging in, e.g. sending the credentials via body or query parameters?

Some background: http://blog.portswigger.net/2016/10/exploiting-cors-misconfigurations-for.html

m-mohr:
Changes with regards to how CORS headers are recommended to be sent. This came up in recent discussions in issue #41 and we realized the current recommendation could lead to security issues (although I don't think it could really be exploited with the current implementations). We figured out the current recommendation was based on a not very well written documentation regarding javascript HTTP request option withCredentials in axios / fetch / XHR. We don't need to provide the `Access-Control-Allow-Credentials` header and can set `Access-Control-Allow-Origin` to `*`. This blocks the potential security issues. I also tried to improve the descriptions around CORS a bit and provide an additional example request/response.

Seeing those changes I don't think this is a breaking change as the old implementations still work and are not invalid (but insecure). So it won't need to go through the PSC process, right? The only thing that will break is that v1.0.0 of the JS client won't be able to access v1.0.1 API back-ends, but JS client [v1.0.1](https://github.com/Open-EO/openeo-js-client/pull/38) will be integrated in all dependents even before the API is actually released.

Fixes #41

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -138,6 +138,8 @@ info:
     Source: [https://en.wikipedia.org/wiki/Cross-origin_resource_sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing)
     
     openEO-based back-ends are usually hosted on a different domain / host than the client that is requesting data from the back-end. Therefore most requests to the back-end are blocked by all modern browsers. This leads to the problem that the JavaScript library and any browser-based application can't access back-ends. Therefore, all back-end providers SHOULD support CORS to enable browser-based applications to access back-ends. [CORS is a recommendation of the W3C organization](https://www.w3.org/TR/cors/). The following chapters will explain how back-end providers can implement CORS support.
+
+    **Tip**: Most servers can send the required headers and the responses to the OPTIONS requests automatically for all endpoints. Otherwise you may also use a proxy server to add the headers and OPTIONS responses.
     
     ## CORS headers
     
@@ -146,10 +148,9 @@ info:
     | Name                             | Description                                                  | Example |
     | -------------------------------- | ------------------------------------------------------------ | ------- |
     | Access-Control-Allow-Origin      | Allowed origin for the request, including protocol, host and port or `*` for all origins. It is RECOMMENDED to return the value `*` to allow requests from browser-based implementations such as the Web Editor. | `*` |
-    | Access-Control-Expose-Headers    | Some endpoints send non-safelisted HTTP response headers such as `OpenEO-Identifier` and `OpenEO-Costs`. All headers except `Cache-Control`, `Content-Language`, `Content-Type`, `Expires`, `Last-Modified` and `Pragma` must be listed in this header. Currently, the openEO API requires at least the following headers to be listed: `Location, OpenEO-Identifier, OpenEO-Costs, Link`. | `Location, OpenEO-Identifier, OpenEO-Costs, Link` |
+    | Access-Control-Expose-Headers    | In addition to the white-listed HTTP headers `Cache-Control`, `Content-Language`, `Content-Type`, `Expires`, `Last-Modified` and `Pragma`, some endpoints need to send additional HTTP response headers such as `OpenEO-Identifier`. All headers that should be made available to browser-based clients MUST be listed in this header. Currently, the openEO API requires at least the following headers to be listed: `Link`, `Location`, `OpenEO-Costs` and `OpenEO-Identifier`. | `Link, Location, OpenEO-Costs, OpenEO-Identifier` |
     
     
-    **Tip**: Most server can send the required headers and the responses to the OPTIONS requests globally. Otherwise you may want to use a proxy server to add the headers and OPTIONS responses.
     
     ### Example request and response
     
@@ -175,11 +176,12 @@ info:
     
     ## OPTIONS method
     
-    All endpoints must respond to the `OPTIONS` HTTP method. This is a response for the preflight requests made by web browsers before sending the actual request (e.g. `POST /jobs`). It needs to respond with a status code of `204` and no response body. In addition to the HTTP headers shown in the table above, the following HTTP headers MUST be included with every response to an `OPTIONS` request:
+    All endpoints must respond to the `OPTIONS` HTTP method. This is a response for the preflight requests made by web browsers before sending the actual request (e.g. `POST /jobs`). It needs to respond with a status code of `204` and no response body.
+    **In addition** to the HTTP headers shown in the table above, the following HTTP headers MUST be included with every response to an `OPTIONS` request:
     
     | Name                             | Description                                                  | Example |
     | -------------------------------- | ------------------------------------------------------------ | ------- |
-    | Access-Control-Allow-Headers     | Comma-separated list of HTTP headers allowed to be send. MUST contain at least `Authorization` if any kind of authorization is implemented by the back-end. | `Authorization, Content-Type` |
+    | Access-Control-Allow-Headers     | Comma-separated list of HTTP headers allowed to be sent with the actual (non-preflight) request. MUST contain at least `Authorization` if any kind of authorization is implemented by the back-end. | `Authorization, Content-Type` |
     | Access-Control-Allow-Methods     | Comma-separated list of HTTP methods allowed to be requested. Back-ends MUST list all implemented HTTP methods for the endpoint. | `OPTIONS, GET, POST, PATCH, PUT, DELETE` |
     | Content-Type                     | SHOULD return the content type delivered by the request that the permission is requested for. | `application/json` |
     
@@ -195,6 +197,8 @@ info:
     Access-Control-Request-Headers: Authorization, Content-Type
     ```
     
+    Note that the `Access-Control-Request-*` headers are automatically attached to the requests by the browsers.
+
     Response:
     
     ```http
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -139,54 +139,72 @@ info:
...
-    | Access-Control-Allow-Origin      | Allowed origin for the request, including protocol, host and port. It is RECOMMENDED to return the value of the request's origin header. If no `Origin` is sent to the back-end CORS headers SHOULD NOT be sent at all. | `http://client.isp.com:80` |
-    | Access-Control-Allow-Credentials | If authorization is implemented by the back-end the value MUST be `true`. | `true` |
-    | Access-Control-Expose-Headers    | Some endpoints send non-safelisted HTTP response headers such as `OpenEO-Identifier` and `OpenEO-Costs`. All headers except `Cache-Control`, `Content-Language`, `Content-Type`, `Expires`, `Last-Modified` and `Pragma` must be listed in this header. Currently, the openEO API requires at least the following headers to be listed: `Location, OpenEO-Identifier, OpenEO-Costs, Link`. | `Location, OpenEO-Identifier, OpenEO-Costs, Link` |
+    | Access-Control-Allow-Headers     | Comma-separated list of HTTP headers allowed to be send. MUST contain at least `Authorization` if any kind of authorization is implemented by the back-end. | `Authorization, Content-Type` |
```
soxofaan, line 182: > Comma-separated list of HTTP headers allowed to be send.

Comma-separated list of HTTP headers allowed to be sent with the actual (non-preflight) request.

openapi.yaml
```diff
@@ -139,54 +139,72 @@ info:
...
-    | Access-Control-Allow-Headers     | Comma-separated list of HTTP headers allowed to be send. MUST contain at least `Authorization` if authorization is implemented by the back-end. | ` Authorization, Content-Type` |
-    | Access-Control-Allow-Methods     | Comma-separated list of HTTP methods allowed to be requested. Back-ends MUST list all implemented HTTP methods for the endpoint here. | `OPTIONS, GET, POST, PATCH, PUT, DELETE` |
+    | Access-Control-Allow-Origin      | Allowed origin for the request, including protocol, host and port or `*` for all origins. It is RECOMMENDED to return the value `*` to allow requests from browser-based implementations such as the Web Editor. | `*` |
     | Access-Control-Expose-Headers    | Some endpoints send non-safelisted HTTP response headers such as `OpenEO-Identifier` and `OpenEO-Costs`. All headers except `Cache-Control`, `Content-Language`, `Content-Type`, `Expires`, `Last-Modified` and `Pragma` must be listed in this header. Currently, the openEO API requires at least the following headers to be listed: `Location, OpenEO-Identifier, OpenEO-Costs, Link`. | `Location, OpenEO-Identifier, OpenEO-Costs, Link` |
```
soxofaan, line 149: > Some endpoints send non-safelisted HTTP response headers such as `OpenEO-Identifier` and `OpenEO-Costs`.

I know it's the technical term, but "endpoints send non-safelisted" sounds like we are doing insecure things. I would change the wording a bit to e.g.

> Some endpoints send important information through custom HTTP response headers such as `OpenEO-Identifier` and `OpenEO-Costs`.


I also think we can drop this sentence:
> All headers except `Cache-Control`, `Content-Language`, `Content-Type`, `Expires`, `Last-Modified` and `Pragma` must be listed in this header.

It's a general CORS implementation detail that is not very relevant here. 
Especially the "all headers" can be confusing: is it "all possible HTTP headers under the sun", "all relevant headers", ...? 
I think the example is good enough to make it clear what a backend should do.

openapi.yaml
```diff
@@ -139,54 +139,72 @@ info:
...
-    | Content-Type                     | SHOULD return the content type delivered by the request that the permission is requested for. | `application/json` |
+    
+    
+    **Tip**: Most server can send the required headers and the responses to the OPTIONS requests globally. Otherwise you may want to use a proxy server to add the headers and OPTIONS responses.
```
soxofaan, line 152: I'm also not sure what the "globally" means in this context. Does it mean something like "automatically"?

(also minor typo: "Most server" -> "Most servers")

soxofaan: Also: I think this tip is a bit out of place here. Maybe it fits better just before the `## CORS headers` header

openapi.yaml
```diff
@@ -139,54 +139,72 @@ info:
     
     openEO-based back-ends are usually hosted on a different domain / host than the client that is requesting data from the back-end. Therefore most requests to the back-end are blocked by all modern browsers. This leads to the problem that the JavaScript library and any browser-based application can't access back-ends. Therefore, all back-end providers SHOULD support CORS to enable browser-based applications to access back-ends. [CORS is a recommendation of the W3C organization](https://www.w3.org/TR/cors/). The following chapters will explain how back-end providers can implement CORS support.
     
-    ## OPTIONS method
+    ## CORS headers
```
soxofaan, line 142: Maybe change title to "General CORS Headers" to make it easier to refer to this from the OPTIONS section

m-mohr: Hmm, I'd rather not change it to not break existing links to this header.

openapi.yaml
```diff
@@ -139,54 +139,72 @@ info:
...
     Content-Type: application/json
+    Location: https://openeo.cloudprovider.com/api/v1/jobs/abc123
+    OpenEO-Identifier: abc123
     ```
     
-    ## CORS headers
+    ## OPTIONS method
     
-    The following headers MUST be included with every response:
+    All endpoints must respond to the `OPTIONS` HTTP method. This is a response for the preflight requests made by web browsers before sending the actual request (e.g. `POST /jobs`). It needs to respond with a status code of `204` and no response body. In addition to the HTTP headers shown in the table above, the following HTTP headers MUST be included with every response to an `OPTIONS` request:
```
soxofaan, line 178: > In addition to the HTTP headers shown in the table above

"In addition to the general CORS headers shown in the table above."

To make it less likely that readers fail to see that this an addition: put "In addition" in italic, or start a new paragraph for this sentence? (case in point: I just realized we there are two CORS header tables in our spec :smile:  )

openapi.yaml
```diff
@@ -139,54 +139,72 @@ info:
...
+    Response:
+    
+    ```http
+    HTTP/1.1 204 No Content
+    Access-Control-Allow-Origin: *
+    Access-Control-Allow-Methods: OPTIONS, GET, POST, PATCH, PUT, DELETE
+    Access-Control-Allow-Headers: Authorization, Content-Type
+    Access-Control-Expose-Headers: Location, OpenEO-Identifier, OpenEO-Costs, Link
+    Content-Type: application/json
+    ```
```
soxofaan, line 207: Should there also be a section about what browser based clients are expected to to do CORS-wise? Like explain the `Access-Control-Request-Method`, `Access-Control-Request-Headers` headers from the request examples?

openapi.yaml
soxofaan: It's a bit unclear to me, but should the`Access-Control-Allow-Headers` always be sent, or only if the OPTIONS request has a `Access-Control-Request-Headers` header?

openapi.yaml
m-mohr: Moved it and replaced "globally" with "automatically for all endpoints".

openapi.yaml
m-mohr: Done

openapi.yaml
m-mohr: That is done automatically by the browsers, I don't actually need to that in the JS client. Thus I added "Note that the `Access-Control-Request-*` headers are automatically attached to the request by the browsers."

openapi.yaml
m-mohr: 👍

openapi.yaml
m-mohr: Hmm, I don't fully agree with this.

> through custom HTTP response headers

Location for example is not really a "custom" HTTP header for me.

Also, I'd personally appreciate if somebody tells me which headers not to list.

Anyway, I tried to improve the situation a bit, e.g. by using white-listed instead of non-safelisted and explaining that "All headers that should be made available to browser-based clients MUST be listed in this header."

m-mohr:
@soxofaan I've pushed changes to the PR, which should tackle (most of) your points.

openapi.yaml
soxofaan: ok, good point.

However, I think this sentence is now bit weird:
> In addition to the white-listed HTTP headers ..., some endpoints need to send additional HTTP ...

I would keep the "some endpoints" statement separate from the white list (which is a global thing).

suggestion:

>  Some endpoints responses include HTTP headers such as `OpenEO-Identifier` (e.g. containing id of a newly created job) and `Location` (for forwarding). To make these headers available to browser based clients, they MUST be white-listed through the `Access-Control-Expose-Headers` response header. (The following headers are white-listed by default and should not be included: `Cache-Control`, `Content-Language`, `Content-Length`, `Content-Type`, `Expires`, `Last-Modified` and `Pragma`.) This version of the openEO API requires at least the following headers to be listed: `Link`, `Location`, `OpenEO-Costs` and `OpenEO-Identifier`.

Note: 
- from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers I learned that `Content-Length` is also whitelisted by default and I added it to the list. Feel free to revert if that is wrong
- I also changed `Currently,` to `This version of the api` to be a bit more explicit version-wise

m-mohr: Good catch on Content-Length! My source didn't list it. Interestingly, it is missing in several sources as it was added later: https://github.com/whatwg/fetch/pull/626

@soxofaan I made the suggested changes, but tried to be a bit more compact. Is this ready to be merged now?

m-mohr:
If there are no other comments, I'd like to merge this and then get the whole API version 1.0.1 through the PSC process instead of doing it for this PR only.

openapi.yaml
soxofaan: yes, looks lot better now

soxofaan (approved):

Issue #310, opened 2020-07-09
Unification of /processes and /process_graphs


soxofaan:
This is follow up issue of #305/#309, which added a `namespace` property to process graph nodes to identify whether the `process_id` property refers to a predefined process or a user defined process.

Possible next steps:
-   merge GET /processes and GET /process_graphs and work with an additional namespace request parameter to select PDP or UDP (or even other namespaces)
-   unify all /processes and /process_graphs requests to something like /processes/{namespace}/{process_graph_id}

jdries:
This issue is becoming relevant again for us: we'd like to allow users to expose their process graphs. This is supported by the namespacing, but it seems they are not yet discoverable?
I would try to avoid breaking the API for this, at least not in a 1.1 version. So certainly the /processes endpoint should not change in an incompatible way.

So I'm proposing something like:
/namespaces -> list all namespaces
/namespaces/{namespace} -> list all processes in a namespace
/namespaces/{namespace}/{process_graph_id} -> list process metadata
This resembles the GitHub approach: our namespaces would be called an 'organization' in GitHub. Users could store process graphs in multiple organizations (namespaces).

By not reusing /processes or / process_graphs we are fully backwards compatible. I'm not yet sure if 'namespaces' is the best name?

jdries:
Rethinking this: the use case of sharing user defined processes is also what the 'marketplace' in openEO platform is supposed to do.
So we could also consider defining an entirely separate microservice, that can move at its own pace, so we can move at a faster pace compared to openEO versions? (Meaning in the beginning, when we are defining and changing such a service.)

aljacob:
What do you mean @jdries  with a seperate microservice? Just an arbitrary http service that is able to work with openEO processes or process_graph and store and offer them? I think the namespaces proposal from you, from yesterday was not bad. We can still keep that as an experimental endpoint for now, without a new API release. 

Thinking more about it...
I have to say I am also a bit confused by the duality of /processes and /process_graphs now that I have implemented version 1. As process_graphs are anyways always wrapped in a process when send to the back-end for processing, maybe it would actually make sense, to organize also their storage and sharing via processes. Similar to what @soxofaan has proposed originally.

m-mohr:
[...]

PR #348, opened 2020-12-07
Process unification #310

m-mohr:
This is a first draft for #310 that tries to unify /processes and /process_graphs.

Sharing might be a separate PR.


openapi.yaml
```diff
@@ -5053,10 +5177,13 @@ components:
...
-      description: >-
+      description: |-
         Unique identifier of the user. MUST match the specified pattern.
-        This is usually a randomly generated internal identifier from the provider
-        not meant for displaying purposes.
+
+        For ease of use, it is NOT RECOMMENDED to use long randomly generated
+        identifiers. More readable user identifiers like `john_doe` support a
+        better user-experience as the user identifier is used in URIs for shared
+        processes, e.g. `https://example.org/api/v1.0/processes/@john_doe/my_ndvi`.
```
soxofaan, line 5186: I agree that `john_doe` is nicer, but this recommendation does not play well with what  OIDC gives us. For example google returns user ids like `us-a44d63b6-e090-6059-81d7-cbe7afeff6ce` and microsoft: `nIrHDS4rhk4ri738TRhtLHXdoUQ6OxZo9Ob0AS3vTig`

there is a piece of the puzzle missing here

openapi.yaml
```diff
@@ -1525,6 +1525,21 @@ paths:
...
+                      A list of namespaces the user has access too.
+
+                      The following namespaces are usually listed:
+                      * `backend`
+                      * `user`: only if authenticated
+                      * `@` + other user id: A list of namespaces of *other* users
+                        which share at least one process accessible by the user.
+                        The namespace with the authenticated user id MUST NOT be listed.
+
+                      For non-authenticated users public processes are taken into account.
```
soxofaan, line 1540: How are "public" processes defined at this point? Is that just pre-defined processes?

openapi.yaml
```diff
@@ -2032,82 +2228,29 @@ paths:
   /process_graphs:
     get:
       summary: List all user-defined processes
-      operationId: list-custom-processes
+      operationId: list-custom-processes-legacy
+      deprecated: true
       description: |-
-        Lists all user-defined processes (process graphs) of the
+        Redirects to all user-defined processes of the
```
soxofaan, line 2234: Also add some kind of deprecation notice to the description of these endpoints?

openapi.yaml
```diff
@@ -4783,10 +4888,29 @@ components:
...
+        namespace for each individual user. They are prescribed with a leading
+        `@` followed by the unique identifier of a user.
+
+        The namespace `user` is an alias for `@` + user id of the authenticated user.
+        It can be used to retrieve the user-defined processes of the authenticated user.
+        
+        The namespace `backend` is an alias for predefined processes.
+
+        Back-end implementations MAY implement other namespaces that don't
+        conflict with any of the namespaces mentioned above.
```
soxofaan, line 4913: see discussion at https://github.com/Open-EO/openeo-api/issues/310#issuecomment-740567496

I think the namespace "format" should be more flexible/generic than: `@` is for per-user namespaces

soxofaan (commented):
(initial review)

openapi.yaml
```diff
@@ -1617,6 +1632,187 @@ paths:
...
+        It is **strongly RECOMMENDED** to keep the response size small by
+        omitting larger optional values from the objects in `processes`
+        (e.g. the `exceptions`, `examples` and `links` properties).
+        To get the full metadata for a user-defined process clients MUST
+        request `GET /processes/{namespace}/{process_id}`.
+      tags:
+        - User-Defined Processes
+        - Process Discovery
+      security:
+        - Bearer: []
```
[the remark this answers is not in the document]
m-mohr: Yes, indeed. I need to fix that.

openapi.yaml
m-mohr: If there's no user name / nick name to derive such a slug from, what to use instead?

openapi.yaml
m-mohr: This mixes the current state and sharing a bit, so if you have a public: true/false flag for sharing, all public processes would be returned.

openapi.yaml
m-mohr: At least it should probably describe what to use/do instead. But I'm still not sure whether to use this redirect behavior (breaking) or leave /process_graphs as it is (basically as an alias, but non-breaking).

openapi.yaml
soxofaan: indeed, I'm not even sure the returned userid is stable, e.g. if a user registers a new client id to work with, maybe an OIDC providers could bump/rehash the userid?

openapi.yaml
soxofaan: ok so this to be compatible with the still to be discussed sharing aspect?

openapi.yaml
soxofaan: how do you mean that the redirect is breaking? That (some) clients don't properly handle a redirect on POST/PUT/PATCH/DELETE?

openapi.yaml
m-mohr: But in general, I would expect a back-end assigns a separate user-id in addition to the external user-id. That is how I've seen it implemented.

openapi.yaml
m-mohr: And I disagree in #310 ;-)

openapi.yaml
m-mohr: Yes, I assume that not all clients would handle it properly. But if we test JS, R and Python and all work flawlessly, it's probably okay to call it non-breaking ;-)

m-mohr:
Seems not stable enough for 1.1, so moving to 1.2. Would be good to have an implementation first, too.

openapi.yaml
```diff
@@ -1617,6 +1632,187 @@ paths:
           $ref: '#/components/responses/client_error_auth'
         5XX:
           $ref: '#/components/responses/server_error'
+  '/processes/{namespace}':
+    get:
+      summary: List all user-defined processes in a namespace
```
soxofaan, line 1637: Documentation of `/processes/{namespace}` and `/processes/{namespace}/{process_id}` talk about "user-defined processes in a "namespace", but listing "predefined" processes should also work, right?

e.g. `GET /processes/backend` returns the same as `GET /processes`
and `GET /process/backend/filter_temporal`  would return the metadata of `filter_temporal` process?

m-mohr: I think the thinking was that there would be duplication as you've pointed out. You usually get all details from /processes for pre-defined processes so that this "extension" is mostly for user-defined processes (and was mostly copied from /process_graphs). So we can discuss whether we should remove the "user-defined" here. The backend namespace is somewhat different though as it is read-only...

m-mohr: Yes, I'm fine with that. We can add additional wording that clarifies any specifics for user-defined processes.

m-mohr: Hmm, looking into the OpenAPI file made me remember that user-defined processes and pre-defined processes have a different schema. So indeed the endpoints were only meant to support user-defined processes so far.

m-mohr: So it seems we need to split up into two separate endpoints to make sure the schemas can be applied correctly:
- '/processes/{namespace}': user-defined
- '/processes/backend': predefined

soxofaan: hmm that's unfortunate. So that means there are still quite some, possibly breaking, steps to take to reach the ultimate goal of "process unification"?

m-mohr: Yes, could be. Still waiting for more feedback, not ideal if it's just us commenting here.
The first question of course is what "process unification" means? Right now it's just about API endpoints, but maybe we also want to go deeper into the required fields? The schemas itself are the same, except for required fields...

soxofaan: > we need to split up into two separate endpoints to make sure the schemas can be applied correctly
> Or we don't allow exposing this endpoint as it's already exposed via /processes. 

Both options are equally bad for us, VITO, as we are already using a non-"backend" namespace containing predefined proprietary processes (without a "process_graph"), which would be invalid according to both of these options.

m-mohr: Hmm... maybe we can add a discriminator to GET /processes/{namespace}. If it contains type: `user: true` in the response it applies the user schema to the processes array, otherwise the predefined schema. That would also help clients to know whether they can do non-GET requests on that namespace.

soxofaan: I think part of the current problem with "predefined" vs "user-defined", is that we are lumping together a couple of process concepts, for example:

predefined | user defined
-----------|-----------
live in default namespace | live in "user" namespace
implemented "natively" by backend | implemented through openEO process graph
has no "process_graph" field in metadata | has a "process_graph" field in metadata
public (by default?) | private (by default?)
no public API to add/update/remove  |  created and managed though openEO endpoints
parameters and return values must be declared | parameters and return values are optional

By sticking to this binary division, each with own "schema", we probably make it hard for ourselves to create a clean API in the long term. 
For example, in VITO backend we already have processes that mix properties from both columns, e.g. private, natively implemented processes that live in a namespace that is neither default or per-user. Another example is defining predefined processes through a "process_graph", instead of "natively" (e.g. "ndvi" or "evi").

m-mohr: Please remember that we are just moving things around here to stay compliant with the API v1.x line. We can't change a lot wrt the schemas for example as that usually is a breaking change. So until we go for API v2 we may need to go with some compromises.

soxofaan: sure, I understand, I was just reflecting on the tension between the constraints of the v1.x API and what we are doing in VITO backend (or want to do) with custom namespaces

openapi.yaml
```diff
@@ -4992,36 +5100,55 @@ components:
...
-        * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
+        * `backend`: Uses exclusively the predefined processes listed at `GET /processes`.
+        * `user`: Uses exclusively the user-defined processes listed at `GET /processes/@{user_id}`
+          with `{user_id}` being the user id of the authenticated user. This is an alias for 
+          directly specifying `@{user_id}` as namespace.
 
         If multiple processes with the same identifier exist, Clients SHOULD
         inform the user that it's recommended to select a namespace.
+    process_namespace:
+      type: string
```
m-mohr, line 5128: I'm struggling with that the namespace allows so many characters. It's hard then to convert those into code and back, e.g. in the Web Editor Math Editor. 

```suggestion
      type: string
      pattern: ^@?[\w\-\.~]+$
```

soxofaan: FYI: we started playing with URL based namespaces to allow UDP sharing, as documented here: https://open-eo.github.io/openeo-python-client/cookbook/udp_sharing.html#using-a-public-udp-through-url-based-namespace
To allow that, the namespace pattern has to be more liberal I'm afraid

m-mohr: Then I won't be able to handle them in JS, I'm afraid. What do you need the other characters for? Other characters are a pain to be used in (generated) code. I was actually only implementing this for a request from VITO (see https://github.com/Open-EO/openeo-web-editor/issues/178) so one of the requests will then not be solvable.

soxofaan: I'm not sure we are talking about the same thing or at least I don't completely understand your struggle with namespace characters.
what I am referring to is using a URL as namespace in the JSON representation of a process graph. Like in the example I linked to:
```json
    "to_celsius": {
        "process_id": "fahrenheit_to_celsius",
        "namespace": "https://openeo.vito.be/openeo/1.0/processes/u:johndoe/fahrenheit_to_celsius",
        "arguments": {"fahrenheit": 86}
    }
```

In my current understanding a namespace is just an opaque string, taking one of the predefined values (like `backend` or `user`), or some backend-specific  value/format. I don't think the namespace has to be parsed or encoded/decoded at the level of the web editor (e.g. in Python client you just pass a string to a `namespace` argument, and the client is not aware what namespaces are valid or what their format should be)

m-mohr: This is a misunderstanding, indeed. This restriction is not in the process graph nodes, but in the URL of the /processes/{namespace}/... endpoints, where {namespace} can only be the given characters.

The process graph node namepsaces can still include arbritrary URLs (although not supported by the JS ecosystem and R has no namespace support at all right now).

m-mohr: By the way, why not simply...

```json
    "to_celsius": {
        "process_id": "fahrenheit_to_celsius",
        "namespace": "u:johndoe", // or @johndoe
        "arguments": {"fahrenheit": 86}
    }
```
?

soxofaan: > but in the URL of the /processes/{namespace}/... endpoints, where {namespace} can only be the given characters.

ok then indeed a restricted character set makes sense. I'm fine with alphanumeric and a couple of special characters that can act as some kind of separator. I don't have strong preference.

m-mohr:
Some additional thoughts on namespaces:

1. The namespaces list in /processes just lists the identifiers. For some use-cases it could be useful that this is expansible, e.g. with a `title`, a `description`, a boolean `loadByDefault` (I guess mostly for the Web Editor?) or a `type` (e.g. [predefined or user-defined](https://github.com/Open-EO/openeo-api/pull/348#discussion_r633286996))
2. Requesting the processes for a namespace could also include the additional metadata for the namespace as mentioned above (excluding `loadByDefault` I guess).

soxofaan:
> Final unification would then be done in v2.

I'm afraid I have to agree :smile:  . 
It's probably not feasible to achieve unification in a clean, non-breaking way under v1

openapi.yaml
```diff
@@ -5081,10 +5195,30 @@ components:
...
-        * `user`: Uses exclusively the user-defined processes listed at `GET /process_graphs`.
+        * `user`: Uses exclusively the user-defined processes listed at `GET /processes/@{user_id}`
+          with `{user_id}` being the user id of the authenticated user. This is an alias for 
+          directly specifying `@{user_id}` as namespace.
 
         If multiple processes with the same identifier exist, Clients SHOULD
         inform the user that it's recommended to select a namespace.
+    process_namespace:
+      type: string
+      pattern: ^@?[\w\-\.~]+$
```
m-mohr, line 5206: For VITO we'd need to add a double colon here.

openapi.yaml
m-mohr: In the meantime I'm fine with more flexibility, but on the other hand we should likely restrict the allowed characters:
https://github.com/Open-EO/openeo-api/issues/478

Issue #349, opened 2020-12-08
Handling of multitemporal cubes with GeoTIFF 


soxofaan:
(I'm not sure this is openeo-api or openeo-processes discussion, so feel free to move.)

The VITO backend provides the GeoTIFF output format for downloading data cubes.
When the data cube spans multiple timestamps, we silently collapse the temporal dimension so that user gets a single date geotiff (possibly having multiple bands). In current implementation, we collapse the temporal dimension by randomly picking a date, which makes the result non-deterministic and which we want to improve (e.g. take last date).

The question now is: should the openeo API prescribe how to handle this problem: data cube has combination of dimensions that the `save_result` format does not support?
The easy answer is of course to throw an error when there is mismatch: the user is responsible for manipulating the cube into correct format before exporting.

But in practice, this auto-reformatting "feature"  makes working with openeo less annoying for a user that quickly wants to inspect a cube they are working with, without having to worry about details (for GeoTIFF I should make sure that ..., for PNG I should make sure that ....)

m-mohr:
Some related issues: 
- https://github.com/Open-EO/openeo-processes/issues/59
- https://github.com/Open-EO/openeo.org/issues/8

But nothing really materialized until now.

@soxofaan Why do you collapse the data and don't return a GeoTiff for each timestamp for example?

soxofaan:
> Why do you collapse the data and don't return a GeoTiff for each timestamp for example

I think because at the time this was implemented we only worked with single file downloads or the spec only provided that, I'm not sure, that was before I joined the project.

jdries:
> Why do you collapse the data and don't return a GeoTiff for each timestamp for example?

Can work for batch jobs, but what about synchronous requests? Can you return multiple geotiffs?

m-mohr:
@jdries I think in theory nothing prevents a back-end from returning a [multipart](https://www.iana.org/assignments/media-types/media-types.xhtml#multipart) response in a sync request, although it's likely that clients do not support that yet.

Of course this issue also arises if you want to export such a process to a web service like a XYZ or TMS.  Strictly, it should probably rise an error. Nevertheless, it may be more user-friendly to define a behavior (e.g. last image) and print a warning into the logs (GEE does that at the moment).

m-mohr:
ToDo: Experiment with and if it all works out nicely clarify in the API that in sync mode multiple files can be returned via multipart responses.

jonathom:
[...]

PR #364, opened 2021-02-24
Multiple files returned by sync. processing #349

m-mohr:
Clarified that `POST /result` may return multiple files as a tar. #349

This is potentially already allowed, at least there's no direct exclusion of the multipart media types, but none of the clients seem to really handle it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `GET /processes` and `GET` / `PUT` for `/process_graphs/{process_graph_id}`: Allow specifying the return values processes receive from child processes. [#350](https://github.com/Open-EO/openeo-api/issues/350)
+- Recommendation that `POST /result` returns a `tar` file if the result consists of multiple files. [#349](https://github.com/Open-EO/openeo-api/issues/349)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


m-mohr:
Please note that the tar recommendation is only for synchronous processing where we basically can only return a single binary blob. This doesn't affect batch jobs or secondary web services. Also, it's just a recommendation. Back-ends can still throw an error. 

If there are no objections to this PR in the next two weeks, I'd merge it as it helps with a number of cases that are not yet specified in the API for sync. processing. Otherwise, please propose an alternative in a PR.

soxofaan:
any reason to pick tar instead of other archive format like zip, gzip or bzip?

(I'm not against tar, just checking whether tar is the best/easiest solution for backends and clients)

m-mohr:
@soxofaan tar is relatively easy to implement as it basically just concatenates files without the client having to implement various sorts of compressions. We basically just want to combine files and that's what tar does. Compression can be done through HTTP compression in gzip or so. For example, the tar js implementation is about 300 lines of code, while the zip implementations have thousands of lines. Same experience earlier in my career with PHP. I guess for the CLI libraries R and Python it's not a big deal whether they run tar or zip, right?

soxofaan:
Python has tar/zip/gzip/bzip support out of the box in the standard library, so no problem there. 
On MacOS and a typical Linux desktop distro, you have support for tar/gzip and zip out of the box too.

But I was indeed wondering about tar support on windows. The client libraries should cover that in most cases.

PR #366, opened 2021-03-17
Add `default_client_id` to OpenID Connect provider data

soxofaan:
First rough attempt to let backend pre-define a default OIDC client id, that openEO clients can use to streamline the OIDC authentication flow

discussion points and possible improvements:
- define a single `default_client_id` or allow multiple? I don't think there is a use case for multiple.
- other relevant information of OIDC client that could be relevant to openEO client? Most obvious example is  redirect URL whitelist for Authorization Code Grant.
- more instructions/constraints for backend provider about how to set up client?

to do:
- add changelog entry

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1831,6 +1831,8 @@ paths:
 
                             The default OpenID Connect client is provided without availability guarantees.
                             The backend implementer CAN revoke, reset or update it any time.
+                            As such, openEO clients SHOULD NOT store or cache default OpenID Connect client information
+                            for long term usage.
                             The default OpenID Connect client is intended to simplify authentication for novice users.
                             For production use cases, it is RECOMMENDED to set up a dedicated OpenID Connect client.
                           required:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1818,29 +1818,56 @@ paths:
 
                             [CommonMark 0.29](http://commonmark.org/) syntax MAY be used for rich
                             text representation.
-                        default_client:
-                          title: Default OpenID Connect Client
-                          type: object
+                        default_clients:
+                          title: Default OpenID Connect Clients
+                          type: array
                           description: |-
-                            Default OpenID Connect client that can be used for OpenID Connect authentication.
+                            List of default OpenID Connect clients that can be used by an openEO client
+                            for OpenID Connect based authentication.
 
-                            The default OpenID Connect client is managed by the backend implementer.
+                            A default OpenID Connect client is managed by the backend implementer.
                             It MUST be configured to be usable without a client secret,
                             which limits its applicability to OpenID Connect grant types like
                             "Authorization Code Grant with PKCE" and "Device flow with PKCE"
 
-                            The default OpenID Connect client is provided without availability guarantees.
+                            A default OpenID Connect client is provided without availability guarantees.
                             The backend implementer CAN revoke, reset or update it any time.
                             As such, openEO clients SHOULD NOT store or cache default OpenID Connect client information
                             for long term usage.
-                            The default OpenID Connect client is intended to simplify authentication for novice users.
+                            A default OpenID Connect client is intended to simplify authentication for novice users.
                             For production use cases, it is RECOMMENDED to set up a dedicated OpenID Connect client.
-                          required:
-                            - id
-                          properties:
-                            id:
-                              type: string
-                              description: OpenID Connect client id
+                          items:
+                            title: Default OpenID Connect Client
+                            type: object
+                            required:
+                              - id
+                              - grant_types
+                            properties:
+                              id:
+                                type: string
+                                description: OpenID Connect client id
+                              grant_types:
+                                type: array
+                                description: |-
+                                  List of authorization flows (grant types) supported by the OpenID Connect client.
+                                  A flow descriptor consist of a OAuth 2.0 grant type,
+                                  with an additional `+pkce` suffix when the flow should be used with the PKCE extension.
+                                minItems: 1
+                                items:
+                                  type: string
+                                  enum:
+                                    - 'implicit'
+                                    - 'authorization_code+pkce'
+                                    - 'urn:ietf:params:oauth:grant-type:device_code+pkce'
+                                    - 'refresh_token'
+                              redirect_urls:
+                                type: array
+                                description: |-
+                                  List of redirect URLs that are whitelisted by the OpenID Connect client.
+                                  Redirect URLs MUST be provided when the OpenID Connect client supports
+                                  the `implicit` or `authorization_code+pkce` authorization flows.
+                                items:
+                                  type: string
                         links:
                           type: array
                           description: |-
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1828,7 +1828,7 @@ paths:
                             A default OpenID Connect client is managed by the backend implementer.
                             It MUST be configured to be usable without a client secret,
                             which limits its applicability to OpenID Connect grant types like
-                            "Authorization Code Grant with PKCE" and "Device flow with PKCE"
+                            "Authorization Code Grant with PKCE" and "Device Authorization Grant with PKCE"
 
                             A default OpenID Connect client is provided without availability guarantees.
                             The backend implementer CAN revoke, reset or update it any time.
@@ -1836,6 +1836,7 @@ paths:
                             for long term usage.
                             A default OpenID Connect client is intended to simplify authentication for novice users.
                             For production use cases, it is RECOMMENDED to set up a dedicated OpenID Connect client.
+                          uniqueItems: true
                           items:
                             title: Default OpenID Connect Client
                             type: object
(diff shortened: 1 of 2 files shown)


m-mohr:
Yeah, this certainly needs a bit more discussion and some experiments. So I'll convert this to a draft for now.

soxofaan:
> other relevant information of OIDC client that could be relevant to openEO client?

another thing that could be relevant: whether the OIDC provider supports device code flow + PKCE, or more generally: for which grant types the OIDC provider supports PKCE

m-mohr:
@soxofaan In theory that should be handled by OIDC discovery (`/.well-known/openid-configuration`), I think. We should try to avoid replicating and making our own OIDC discovery as far as possible. So the first thing to figure out is whether any of this is exposed by the OIDC discovery, right?

soxofaan:
> that should be handled by OIDC discovery

Indeed, I was hoping for that too, but so far I could not find any standard way of specifying PKCE features in the discovery document.

e.g. for EGI Check-in (https://aai.egi.eu/oidc/.well-known/openid-configuration), this is the only bit related to PKCE:
```json
  "code_challenge_methods_supported": [
    "plain",
    "S256"
  ],
```
But that does not refer to device code + PKCE specifically.

In Google OIDC discovery https://accounts.google.com/.well-known/openid-configuration you have the same bit, but device code + PKCE does not seem to be supported (based on initial testing

m-mohr:
Yes, I also have not found anything of interest. It's probably because PKCE was written with only AuthCode in mind and Device Flow + PKCE is not even standardized (at least I can't find it).
The default_client_id only works for Device + PKCE, right? For AuthCode + PKCE or Implicit you need a redirect URL, all others need a secret. So it's a bit of an unstandardized behavior we'd add to the spec so I'm just a bit concerned about it.

soxofaan:
> The default_client_id only works for Device + PKCE, right? For AuthCode + PKCE or Implicit you need a redirect URL, all others need a secret.

The default client will probably be most useful for Device+PKCE.
Although in principle it could also be usable for AuthCode+PKCE e.g. when the redirect URL whitelist contains some well chosen localhost urls, which could help users working on localhost apps. Or maybe even some (localhost) Jupyter use cases.

soxofaan:
> So it's a bit of an unstandardized "niche" behavior we'd add to the spec so I'm just a bit concerned about it.

It's indeed a bit "niche", but without it I'm afraid that the user experience of authentication with the openeo python client will be so poor (for non-expert user) that we probably end up promoting basic auth or risk losing users before they even can try a thing.

You could also see it the other way: the fact that is unstandardized in general OIDC specs is actually a good reason to "standardize" a bit in the openeo API :stuck_out_tongue:

soxofaan:
To have a bit more future proof wiggle room I changed it to a `default_client` object with a single `id` field (for now).
Feel free to fine tune the description

m-mohr:
If we go this route to expose client_ids via API, we should also make it work for Jupyter + Web Editor. I'll post an example later.

soxofaan:
In Open-EO/openeo-python-client#192  I've already implemented support for using the default client as proposed here:

Screenshot from 2021-03-19 11-02-32

-> no client_id or secret to specify :)

m-mohr:
What do you think about something like this, @soxofaan ?

```
{
  "clients": [
    {
      "id": "abcd",
      "flow": "authorization_code",
      "pkce": true,
      "redirect_urls": [
        "https://localhost/callback",
        "https://editor.openeo.org"
      ]
    },
    {
      "id": "edgh",
      "flow": "implicit",
      "redirect_urls": [
        "https://localhost/callback",
        "https://editor.openeo.org"
      ]
    },
    {
      "id": "ijkl",
      "flow": "device",
      "pkce": true
    }
  ]
}
```

soxofaan:
I didn't want to go as far as making it possible to specify multiple default clients, but it could indeed be necessary for OIDC providers that don't allow to create clients that cover multiple flows. 
Did you intentionally name the top level field "client" instead of "default_clients"? I think it should be clear that the list is a list of fallback/default options, not a white list of allowed clients.

> We may be able to optimize this, e.g. putting pkce into the flows ad making flows an array, e.g. flows: ['implicit', 'device+pkce']

Yeah, that would make things more compact for most use cases (like EGI Check-in or Keycloack, which allow to define clients that support multiple flows).

m-mohr:
An open question is what values would be allowed for `flow`. Is there a list that is standardized? Otherwise, we need to come up with our own list of values and that would of course not be as extensible as a list of values that comes from OIDC directly. But those lists will have the issue that device flow with or without pkce are not standardized in there. So what potential values are there?
- implicit
- device+pkce
- authorization_code+pkce
- ...?

m-mohr:
I just found:

> grant_types_supported
> OPTIONAL. JSON array containing a list of the OAuth 2.0 Grant Type values that this OP supports. Dynamic OpenID Providers  MUST support the authorization_code and implicit Grant Type values and MAY support other Grant Types. If omitted, the default value is ["authorization_code", "implicit"].

Source: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

Does that imply that if grant_types is not listing device flow in the Google and EGI documents, that are officially not supported?

soxofaan:
> So this workaround/specification is mostly to fix that device+pkce is not standardized?

- If you mean the whole PR is to fix that device+pkce is not standardized: no, the core of the main point of the PR is to let a backend advertise default OIDC client ids to eliminate that challenge  for the less technical user.
- If you mean just the the part about `grant_types`: yes, it is unfortunate that we can not reliably detect supported grant types with PKCE support from the OIDC discover document directly


About the JSON code snippet: I think that starts to look good. I'll update the PR to this format

> Either say in the description that only values allowed for grant_type_supported can be used. Alternative: Specify an enum.

Although I wouldn't over-specify this, it's probably best to work with enum

> Say that if used all except implicit must implement PKCE or define a property that specifies use of pkce per grant type

I don't think we have to _require_ usage of  PKCE explicitly this way. The main point should be that the client can be used without a client secret. PKCE is just one semi-standardized to do that. The Implicit flow is another example that does not require a secret, but also does not use PKCE. 
In that sense, I think we should indeed include that `+pkce` suffix that you suggester earlier, so that the clients can act properly. The grant  type enum list should be then as follows I think:
- `implicit`
- `authorization_code+pkce`
- `urn:ietf:params:oauth:grant-type:device_code+pkce`
- `refresh_token`

m-mohr:
Okay, that all sounds good to me. Let me know in case you need any help. I can also take over the specification work, if you are busy with other things right now.
 
Do we really need to add refresh_token to the enum?

soxofaan:
> Do we really need to add refresh_token to the enum?

Technically yes, because it is also a grant type. And it will be used in practice. So adding it makes things more consistent.

m-mohr:
I don't agree with it yet. You don't need refresh_token to connect, that's just just afterwards to refresh token. But at that point you are already gone through the OIDC discovery, where this is listed already. So what's the point of listing it? I'd understand it that I can authenticate with just a client_id and the refresh_token flow, but that is not the case, right?

soxofaan:
> You don't need/use the refresh_token flow to connect, that's just just afterwards to refresh token. But at that point you've already gone through the OIDC discovery, where this is listed already. 

I'm not following completely, but I think you are reasoning in a interactive/web app context here.
In the context of python scripts, the refresh token flow is a lot like other flows. For example: you run a script, authenticating with device code flow (and storing refresh token in background). One week later you run the script again, using the refresh token directly, as if were the "initial" authentication step.

> I'd understand that I can authenticate with just a client_id and the refresh_token flow, but that is not the case, right?

That's actually one of the aims for non-interactive openeo scripts (python or other). And this PR helps with making that goal accessible for less technical users.

The python client lib has now an experimental new method  `Connection.authenticate_oidc()` that first tries the refresh token flow, and if that fails, it falls back on device code flow. So now I put just this in my openeo notebooks against the VITO backend:

     con.authenticate_oidc(provider_id="egi")

and most of the times it just immediately works without any interaction using the refresh token flow. And very rarely when it fails (e.g. no or expired refresh token) it starts the device code flow. This makes a very clean user experience.

m-mohr:
I understood this, but my point is that if you already have a refresh token, you certainly know that this is supported and don't need it in discovery. It don't think we need to expose it as the default client option is just needed for the "first" authentication. For the refresh token you don't need a default client id and that's what this is about, right?

soxofaan:
> Let me know in case you need any help. I can also take over the specification work,

just pushed an update, feel free to take over and do finetuning

soxofaan:
> For the refresh token you don't need a default client id and that's what this PR is about, right? But maybe I'm wrong and refresh token flow needs to specify a client id again?

As far as I have experienced with the providers I have tried, the client id must indeed be included with the request where you exchange the refresh token for a new access token.
Because the refresh token flow is a lot like the other flows (in terms of token request fields, configuration on OIDC provider backend, etc), it feels more consistent to include it in that list. 
I understand that there is duplication here with what can be found in the OIDC discovery document, but that's the unfortunate context of this PR: bridging the gap of under-specified corners of OIDC.

m-mohr:
If that's the case, then it should be included. I wasn't aware a client id is required for the flow.

m-mohr:
@soxofaan I've made some minor changes, a review of commit https://github.com/Open-EO/openeo-api/pull/366/commits/6a8f57240c89be9c60c3338df837c4a8d6de3abe would be great. Then we are ready to merge, I think.

Somewhat related: Would it make sense to allow specifying a default OIDC provider?

m-mohr (approved):

openapi.yaml
```diff
@@ -1828,14 +1828,15 @@ paths:
                             A default OpenID Connect client is managed by the backend implementer.
                             It MUST be configured to be usable without a client secret,
                             which limits its applicability to OpenID Connect grant types like
-                            "Authorization Code Grant with PKCE" and "Device flow with PKCE"
+                            "Authorization Code Grant with PKCE" and "Device Authorization Grant with PKCE"
```
soxofaan, line 1831: I haven't seen usage of "Device Authorization Grant" outside of RFC 8628. These are more common in the wild I think:
- Device Code Grant
- Device Flow

soxofaan:
> Somewhat related: Would it make sense to allow specifying a default OIDC provider?

I have also thought about that and it would be nice improvement. But let's move that to another PR?

openapi.yaml
m-mohr: In the spec, I'd prefer the official language. Additionally, below it is clarified: 

> Device Authorization Grant (aka Device Code Flow) as specified in [RFC 8628](https://www.rfc-editor.org/rfc/rfc8628.html), with PKCE extension.

That should be clear. Other documentation can surely use more common terminology.

openapi.yaml
```diff
@@ -1818,6 +1818,68 @@ paths:
...
+                                description: |-
+                                  List of authorization grant types (flows) supported by the OpenID Connect client.
+                                  A grant type descriptor consist of a OAuth 2.0 grant type,
+                                  with an additional `+pkce` suffix when the grant type should be used with
+                                  the PKCE extension as defined in [RFC 7636](https://www.rfc-editor.org/rfc/rfc7636.html).
+
+                                  Allowed values:
+                                  - `implicit`: Implicit Grant as specified in [RFC 6749, sec. 1.3.2](https://www.rfc-editor.org/rfc/rfc6749.html#section-1.3.2)
+                                  - `authorization_code+pkce`: Authorization Code Grant as specified in [RFC 6749, sec. 1.3.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-1.3.1), with PKCE extension.
+                                  - `urn:ietf:params:oauth:grant-type:device_code+pkce`: Device Authorization Grant (aka Device Code Flow) as specified in [RFC 8628](https://www.rfc-editor.org/rfc/rfc8628.html), with PKCE extension.
```
m-mohr, line 1862:
```suggestion
                                  - `urn:ietf:params:oauth:grant-type:device_code+pkce`: Device Authorization Grant (aka Device Code Flow) as specified in [RFC 8628](https://www.rfc-editor.org/rfc/rfc8628.html), with PKCE extension. Note that the combination of this grant with PKCE is *not standardized* yet.
```

m-mohr: Ugh, wrong line.

PR #367, opened 2021-03-17
Make sure to use "OpenID Connect" in full instead of "OpenID"

soxofaan:
It could be intentional, but I noticed that the docs use "OpenID provider" instead of "OpenID Connect provider".
Technically and practically these are not the same. I think we should use "OpenID Connect" in full wherever possible


m-mohr:
Agreed, this was meant to be "OpenID Connect" always. Not sure why it was written differently. Thanks for fixing it.

Issue #371, opened 2021-04-07
Better wording about case insensitive handling


soxofaan:
(from Open-EO/openeo-processes-python#22)

https://github.com/Open-EO/openeo-api/blob/81dd248b9e8b8caeedbdd96c6a9dd8f3f56d0eb7/openapi.yaml#L974

Wording is bit confusing (sound like backend can choose to be case insensitive or not).

Proposal:

> Format names MUST be handled case insensitively throughout the API.

Same applies to `Service names` and `Billing plans`

m-mohr:
Just made a change, but my git is acting up and I can't commit yet. Will be available "soon".

m-mohr:
> Format names MUST be accepted *case insensitive* throughout the API.

m-mohr:
See PR #372

PR #372, opened 2021-04-07
Clarified case-insensitivity

m-mohr:
Clarified that billing plans, service names and file formats must be accepted case-insensitive. #371

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -728,8 +728,8 @@ paths:
                             name:
                               type: string
                               description: >-
-                                Name of the plan. It MUST be accepted *case
-                                insensitive* throughout the API.
+                                Name of the plan. It MUST be accepted in a *case
+                                insensitive* manner throughout the API.
                               example: free
                             description:
                               type: string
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -729,7 +728,7 @@ paths:
                             name:
                               type: string
                               description: >-
-                                Name of the plan. Is allowed to be *case
+                                Name of the plan. It MUST be accepted *case
```
soxofaan, line 731: I think this is not completely correct grammatically. The "case sensitive" must become an adverb, but "must be accepted case insensitively" sounds also bit weird. Do we have native English speakers that can we consult on this?

m-mohr: Hmm, not sure. Maybe:

> Format names MUST be accepted in a case-insensitive manner throughout the API.

soxofaan: yeah it's probably safer to workaround it that way :smile:

m-mohr: Changed.

soxofaan (approved):

PR #373, opened 2021-04-07
Support a default provider for OIDC + new example

m-mohr:
- Support a default client for OIDC
- Added a new example, also showing the default_clients

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `GET /processes` and `GET` / `PUT` for `/process_graphs/{process_graph_id}`: Allow specifying the return values processes receive from child processes. [#350](https://github.com/Open-EO/openeo-api/issues/350)
 - `GET /credentials/oidc` can provide a set of default client ids for OpenID Connect. [#366](https://github.com/Open-EO/openeo-api/pull/366)
+- `GET /credentials/oidc` can specify a default provider for OpenID Connect.
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


soxofaan:
issue title should say "default provider" instead of "default client"?

m-mohr:
Yes, indeed.

soxofaan:
not ideal that you can not enforce that at most one provider can be flagged as default.

Alternatives:
- add a top level field "default_provider"  (type: string) to point to id of the default provider
- instead of "default" boolean flag: add "priority" integer field: highest priority is default provider
- first provider in list should be considered default provider

m-mohr:
Yes, that's always the issue. You can't really provide a good way of doing it consistently. If you do "default_provider", then you can't check whether the value actually exists. Similarly for the priority, where you can't check whether all fields may be 0 for example.
The first in the list semantic could work, but is somewhat giving a meaning to something a provider may not have chosen explicitly as he may have implemented it in 1.0 where this was not the case yet. Let's say it's not really breaking, but also not really the nicest way of being backward-compatible.

m-mohr:
@soxofaan Would you prefer the alternative of the first array element being the default? I think that's the only option that would lead to a bit more consistency.

soxofaan:
Yes, "first of list" would be my preference. It introduces no real API changes, except for documenting this "convention". It would also work for automatically for backends that just specify a single OIDC provider.

Second preference would be new "default_provider" field because if there is an inconsistency then it will fail in a clear way. If you have inconsistency with the boolean "default" flags, then some implementations might still accept that (because they pick the first match they find).

m-mohr:
I'll update this PR to work with the array convention.

m-mohr:
PR has been updated. This really feels much better integrated, it also works out of the box with the Web Editor implementation.

soxofaan (approved):

Issue #341, opened 2020-10-28
Revoke signed URLs


m-mohr:
The batch job results are provided as "signed" URLs. If you share the URLs, you can't revoke access except you delete the results. But maybe you want to keep the results, but just revoke access to the URLs you may have shared with others. Thus it could be useful to have an API endpoint to revoke the old signed URLs and generate new "signed" URLs.

soxofaan:
I searched around a bit to find info or example implementations on revoking signed URLs but the results were disappointing so far. It seems that the most common approach to improve security is "use short expiration time" (order of minutes). The big implementations like AWS and google cloud don't seem to support revoking of signed URLs at all, except for nuclear options along the lines of "remove the resource" or "revoke the permissions of the resource owner".

So I wonder if there are some pitfalls to this feature request so that no backend will implement it in practice.

m-mohr:
Yes, we will likely need to get more experience with this functionality first. It would also be interesting what other functionality may be required for batch job results (and/or the PATCH endpoint) so that we can figure out how to combine it.
The aim was to get a proposal up for experiments, but I doubt we'll release this in 1.1.

PR #381, opened 2021-04-15
Revoke signed URLs #341

m-mohr:
Proposal for issue #341

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `GET /processes` and `GET` / `PUT` for `/process_graphs/{process_graph_id}`: Allow specifying the return values processes receive from child processes. [#350](https://github.com/Open-EO/openeo-api/issues/350)
+- `PATCH /jobs/{job_id}/results` to revoke signed URLs generated for batch job results. [#341](https://github.com/Open-EO/openeo-api/issues/341)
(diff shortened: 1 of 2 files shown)


soxofaan:
Proposed commit looks fine.

some thoughts:
- PATCHing `assets` to `null` might also be used to remove batch job result files from the backend (e.g. to reduce your storage bill without removing batch job metadata)
- What if the backend does not support invalidating urls, e.g. because the (third party) storage solution does not support it? Should the backend then not support `PATCH .../results` (through the capabilities doc), or should it throw an error (e.g.FeatureUnsupported)?

m-mohr:
> * PATCHing `assets` to `null` might also be used to remove batch job result files from the backend (e.g. to reduce your storage bill without removing batch job metadata)

That would still be possible. `assets: null` is not allowed in STAC so we can use it for this special behavior. To actually remove assets, you'd send `assets: {}` (i.e. empty object) to be STAC-compliant.

> * What if the backend does not support invalidating urls, e.g. because the (third party) storage solution does not support it? Should the backend then not support `PATCH .../results` (through the capabilities doc), or should it throw an error (e.g.FeatureUnsupported)?

Endpoints in general should not be listed in the capabilities so that clients don't request them. Whether back-ends return a 501 or 404 or something else in the 4xx or 5xx range, if clients request the endpoint anyway is mostly up to them.

soxofaan:
> assets = null  to revoke all previously   generated signed URLs 
> To actually remove assets, you'd send assets: {}

To be honest, this looks quite cryptic and somewhat ad-hoc to me

m-mohr:
Indeed, looking at it after a while makes it look weird. Maybe a specific HTTP header or a separate HTTP endpoint (although an issue with REST conventions) are better options.

PR #385, opened 2021-04-28
Clarified allowed characters in the `path` for uploaded user files

m-mohr:
Clarified allowed characters in the `path` for uploaded user files, related to a PR I'll issue soon in processes.

Although this is a "new restriction" in the API, I'd argue this is non-breaking as no OS I'm aware of allows those characters anyway and thus files paths containing such characters would be rejected anyway.


soxofaan (approved):
looks good

you might blacklist some more characters (e.g. &, %, *, @, $), but I'm not sure that helps a lot

m-mohr:
Yeah, I looked into what characters to specify here, but it seems some of these characters are allowed in some environments, so I decided to go with a minimum set that is not working in all envs.
I just had in mind solving https://github.com/Open-EO/openeo-processes/pull/245 and aligning with it anyway.

m-mohr:
Would you mind having a quick look at https://github.com/Open-EO/openeo-processes/pull/245, too? If you think that also works for you, then I think we can merge both. I just want them to stay in sync.

PR #389, opened 2021-05-05
Default plan per user

m-mohr:
This specifies a default plan per user, which the user may have subscribed to.


m-mohr:
That sounds all in line with the PR, thanks.

Issue #410, opened 2021-08-20
Support OIDC device code flow without PKCE


m-mohr:
There's the possibility to support Device Code without PKCE it seems.

Proposal is to extend the list of allowed grant types:
https://github.com/Open-EO/openeo-api/blob/master/openapi.yaml#L1859-L1868 
with
`urn:ietf:params:oauth:grant-type:device_code`

cc @soxofaan @aljacob

m-mohr:
It seems like authorization code flow also works without PKCE and client secret, so should be added, too.

soxofaan:
> It seems like authorization code flow also works without PKCE and client secret, so should be added, too.

In what context or with what provider have you observed that?

m-mohr:
@soxofaan None of ours, but @aljacob mentioned it today, and reading the OAuth 2.0 RFC it is indeed standardized without PKCE and PKCE itself is only an extension.

m-mohr:
Interestingly, it seems like the R client supports authorization_code without PKCE, but with secret.

PR #411, opened 2021-08-25
OIDC: Allow auth code flow and device flow for default clients

m-mohr:
Fixes/Implements #410.

Could someone verify that my understanding of the authorization_code flow is correct and doesn't require a client_secret?


aljacob (approved):
thanks Matthias, with we should have the most relevant oidc use cases covered!

soxofaan (approved):

Issue #416, opened 2021-10-18
Clarify: Send token to discovery endpoints


m-mohr:
Discovery endpoints (e.g. `GET /collections`) allow to be requested with and without Bearer Token. 
Sending the Bearer Tokens with the clients is important once you have it so that private data (e.g. Collections) can be requested.
This is important for e.g. exposing Planet data through openEO Platform. 

Clarify in the API that clients should send tokens if available. This will likely result in issues for all clients, none supports this yet AFAIK.

PR #417, opened 2021-10-19
Clarify: Send token to discovery endpoints #416

m-mohr:
See issue #416 and discussions around private collections in openEO Platform.

This is not really a new thing, but more a clarification for clients that this behavior can be useful.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -138,7 +138,7 @@ info:
 
     <SecurityDefinitions />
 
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -138,6 +138,8 @@ info:
 
     <SecurityDefinitions />
 
+    **Note:** Although it is possible to request several public endpoints for capabilities and discovery that don't authorization, it is RECOMMENDED that clients (re-)request the public endpoints that support Bearer authentication with the Bearer token once available to also retrieve any private data that is made available specifically for the authenticated user.
```
soxofaan, line 141: > that don't authorization,

there is a "require" or "need" missing here I think

soxofaan:
> I guess that meant "Should be easy"? 

indeed :)

> Is the Python client sending the Bearer token to Collections etc if available?
python client always sends bearer if available (even if not necessary)

dthiex (approved):

PR #419, opened 2021-10-19
Federation Extension

m-mohr:
A first draft for the federation extension, all up to discussion, it's really just capturing ideas right now.

Rendered version: https://github.com/Open-EO/openeo-api/blob/federation-extension/extensions/federation/README.md

Related issue: https://github.com/openEOPlatform/architecture-docs/issues/188

### DIFF
--- a/extensions/federation/README.md
+++ b/extensions/federation/README.md
@@ -103,17 +103,14 @@ Applies to:
 - `GET /process_graphs`
 - `GET /jobs`
 - `GET /jobs/{job_id}`
-- `GET /jobs/{job_id}/logs`
 - `GET /jobs/{job_id}/results`
 - `GET /services`
-- `GET /services/{service_id}/logs`
 
-The following two endpoints are not extensible and as such the additional field can't be added:
+The following endpoints define the resources (UDF runtimes / service types) at the top level of their response as key-value pairs.
+Consequently, they are not extensible with additional properties for federation purposes.
 
 - `GET /udf_runtimes`
 - `GET /service_types`
-
-*This is due to the fact that these two endpoints return an object with a key-value-pair per UDF runtime / service type and adding a key-value-pair for federation purposes would be interpreted as a new UDF runtime or service type by clients.*
 
 ### OpenAPI fragment
 
--- a/extensions/federation/README.md
+++ b/extensions/federation/README.md
@@ -141,6 +141,16 @@ schema:
       items:
         type: string
         description: The ID of a back-end.
+    'federation:description':
+      description: |-
+        Allows back-ends to provide backend-specific details about this resource in a human-readable way.
+
+        [CommonMark 0.29](http://commonmark.org/) syntax MAY be used for rich text representation.
+      type: array
+      minItems: 1
+      items:
+        type: string
+        description: The ID of a back-end.
 ```
 
 **Note:** In Collections this should be placed inside `summaries` if users should be able to filter on back-ends in `load_collection`.
@@ -162,7 +172,7 @@ schema:
     {
       "name": "proprietary_parameter",
       "description": "A parameter supported only by EODC.",
-      "federation:hosted": ["eodc"],
+      "federation:backends": ["eodc"],
       "schema": {}
     }
   ]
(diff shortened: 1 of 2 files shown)


m-mohr:
From the meeting today: It seems we also need a flag that the user can set for batch jobs/sync proc/web services so that they can choose explicitly which back-end to send a process graph to. Or should this be done via load_collection properties? The latter seems a bit weird.

soxofaan:
> Or should this be done via load_collection properties? The latter seems a bit weird.

indeed, that feature is focused on specifying the backend to use for providing a particular collection, which currently directly implies the back-end to execute the whole graph


> It seems we also need a flag that the user can set for batch jobs/sync proc/web services so that they can choose explicitly which back-end to send a process graph 

Another idea that crossed my mind is having back-end specific connection urls in addition to the general one, e.g. 
- https://openeo.cloud/eodc/openeo/1.0/ 
- https://openeo.cloud/vito/openeo/1.0/
- ...

so you can interact with an explicitly chosen backend, but still use the general  platform features (auth, billing, ...).
Putting a back-end identifier in the base url avoids having to define some kind of selection field in multiple end-points. And you  can easily provide collection/process discovery without all the tricky metadata merging.

m-mohr:
Another note from the meeting today: 
Do we need a way to expose for collections and processes for which execution type (batch, sync, web services) they are (not) available?

I'm not sure we need it here or in the core API?

soxofaan:
> Do we need a way to expose for collections and processes for which execution type (batch, sync, web services) they are (not) available?
> I'm not sure we need it here or in the core API?

sounds like something that is not (only) related to federation

m-mohr:
Added a new issue: https://github.com/Open-EO/openeo-api/issues/429

m-mohr:
> Another idea that crossed my mind is having back-end specific connection urls in addition to the general one [...]
> so you can interact with an explicitly chosen backend, but still use the general platform features (auth, billing, ...). Putting a back-end identifier in the base url avoids having to define some kind of selection field in multiple end-points. And you can easily provide collection/process discovery without all the tricky metadata merging.

Yeah, I see pros and cons for both approaches. I think we need to discuss this in detail and check what is the better way forward.

Some that come to mind right now:

**Back-end URLs:**
Pros: Avoids merging metadata (could be mitigated by federation extension and some hand-drafted metadata)
Cons: You need to reconnect each time


**Flag in requests:**
Pros: You need to connect once (think web editor) and then can select each time easily in the UI (but this is less easy in programming libraries)
Cons: Needs changes in clients

soxofaan:
>  Cons: You need to reconnect each time

I don't think this will a big issue in practice: the choice to work on a specific back-end instead of working with the federated one will be very conscious decision, so it's a good thing to make that clear.
Moreover, a lot of things are impacted from switching between the federated backend and the underlying backends: collections, processes, file formats, udf runtimes, ... (as discussed elsewhere), so I think it's a good thing that this is reflected "UI-wise" by having to create a separate connection.

m-mohr:
Indeed, this might be biased from a Web Editor standpoint although as a user I'd personally still prefer to have a simple switch. This could be something to leave out for now and then ask our users what they would prefer.

extensions/federation/README.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+#### Collection
+
+```json
+{
+  "stac_version": "1.0.0",
+  "id": "example",
+  "description": "...",
+  "federation:description": {
```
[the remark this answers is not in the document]
m-mohr: Removed `federation:description` for now, let's keep it simple and add it once we may need it.

extensions/federation/README.md
```diff
@@ -92,7 +92,28 @@ schema:
...
+- `GET /jobs/{job_id}/results`
+- `GET /services`
+- `GET /services/{service_id}/logs`
+
+The following two endpoints are not extensible and as such the additional field can't be added:
+
+- `GET /udf_runtimes`
+- `GET /service_types`
+
+*This is due to the fact that these two endpoints return an object with a key-value-pair per UDF runtime / service type and adding a key-value-pair for federation purposes would be interpreted as a new UDF runtime or service type by clients.*
```
soxofaan, line 116: > The following two endpoints ...
> ...
> *This is due to the fact ...*


It took me couple of attempts to understand what you explain here.

proposal to simplify:

> The following endpoints define the resources (UDF runtimes / service types) at the top level of their response as key-value pairs. Consequently they are not extensible with additional properties for federation purposes.
> - `GET /udf_runtimes`
> - `GET /service_types`

m-mohr: Thanks, updated.

m-mohr:
Merged for now, let's discuss further issues in separate PRs.

Issue #426, opened 2021-11-02
capabilities billing default_plan


soxofaan:
https://github.com/Open-EO/openeo-api/blob/f303d65a3291d4cd74dacc0e796803bb5d6fa03b/openapi.yaml#L720-L726

first, a small typo:
> ... or has no default plan has been assigned  ...

While thinking about it, it seems this is not well defined yet:
Should the default plan be a paid or non-paid plan?
If it's a paid plan, then how can it be a default for free users?
If it's a non-paid plan, then it's not ideal to use as default for paying users (e.g.  they get worse performance because they are not using their paid plan).

soxofaan:
(related to #425 )

soxofaan:
given #425 :  maybe there should be no global default plan, and just a per user default plan (based on their subscription)?

m-mohr:
I've tried to improve documentation in PR #428.

jdries:
I'm also really not sure that we'll end up with a notion of multiple billing plans per user. 
There's also this vsc4 'plan', which is not really a billing plan but rather an entirely separate backend, which might even have specific capabilities.

m-mohr:
Sure, but this is the general API so the option should be there anyway. Somewhat agree on vsc4. One the one hand it's a somewhat separate back-end, but on the other hand the billing plans were also meant to distinguish between different "configurations of hardware". So depends on a case by case basis.

Issue #425, opened 2021-11-02
/me: default_plan confusion 


soxofaan:
https://github.com/Open-EO/openeo-api/blob/f303d65a3291d4cd74dacc0e796803bb5d6fa03b/openapi.yaml#L3431-L3441

> Name of the plan the user has subscribed to.

Field is called "default_plan" but description says "name of the plan the user has subscribed to, so field should be called "plan" instead of the fallback-sounding "default_plan" ?

> Overrides the default plan of the back-end, but back-ends
>                      MAY also allow overriding this plan for each individual
>                     processing request (e.g. job or service) with the
>                    corresponding `plan` property.

I think this paragraph makes it even more confusing. A default that overrides another default: what's the point of the latter default then? And it also suggests that a user can work with multiple plans, so there should be a listing of this under `/me` as well I guess?

m-mohr:
Here's how a back-end should resolve it:
1. If provided, use the `plan` from the job/web service.
2. If available, use the default_plan for the user.
3. Otherwise, use the default_plan of the back-end.

The API can't know what a model a back-end uses so it allows flexibility here (and point two was only added in v1.1 so point 3 is a bit of a legacy thing).

m-mohr:
> Field is called "default_plan" but description says "name of the plan the user has subscribed to, so field should be called "plan" instead of the fallback-sounding "default_plan" ?

Naming is probably not ideal, but we can't change it anymore so I'll try to improve the situation through documentation, see PR #428. The PR should also tackle the other issue raised.

Issue #374, opened 2021-04-08
2.0.0 clean-up


m-mohr:
We (may have) introduced a number of changes that we made only for backward-compability and they should be removed in 2.0.0. A list of potential things to do:

- [ ] Remove language that required all assets being available in Collection Assets, see #343 and #359
- [ ] Remove /process_graphs/... endpoints, see #310 and #348
- [ ] Remove support for STAC 0.9.0 and 1.0.0-beta/rc and update to STAC API 1.0
- [ ] Require `type` field in Collections (Collections and Batch Job Results as Collections) and Catalogs (`GET /`)

PR #428, opened 2021-11-04
Clarify (default) billing plans

m-mohr:
Fixes the situation around the issues raised in #425 and #426.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3500,14 +3500,7 @@ paths:
                       the user's real name or a nickname.
                   default_plan:
                     type: string
-                    nullable: true
-                    description: |-
-                      Name of the plan the user has subscribed to, if any.
-
-                      Back-ends resolve billing plans in the following order:
-                      1. If provided, use the `plan` in the processing request.
-                      2. If provided, use the plan exposed here.
-                      3. Otherwise, use the `default_plan` exposed through `GET /`.
+                    description: Name of the single plan the user is currently subscribed to if any.
                     example: free
                   storage:
                     title: User Storage
@@ -4085,15 +4078,10 @@ components:
         The billing plan to process and charge the job or service with.
 
         Billing plans MUST be accepted in a *case insensitive* manner.
-        Back-ends MUST resolve the billing plan in the following order:
-        1. Use the `plan` that has been provided in the request, if a non-null value is present.
+        Back-ends MUST resolve the billing plan in the following way if billing is supported:
 
-        If `null` has been provided for the `plan`:
-        2. Use the `default_plan` exposed through `GET /me`, if available.
-        3. Otherwise, use the `default_plan` exposed through `GET /`.
-
-        The resolved plan MUST be persisted permanently, regardless of any 
-        changes to the exposed billing plans in `GET /` in the future.
+        * If a value is given and it is not `null`: Persist the `plan` that has been provided in the request.
+        * Otherwise, don't change the billing plan.
 
         Billing plans not on the list of available plans MUST be rejected with
         openEO error `BillingPlanInvalid`.
@@ -4105,10 +4093,14 @@ components:
         The billing plan to process and charge the job or service with.
 
         Billing plans MUST be accepted in a *case insensitive* manner.
-        Back-ends MUST resolve the billing plan in the following order:
-        1. Use the `plan` that has been provided in the request, if a non-null value is present.
-        2. Use the `default_plan` exposed through `GET /me`, if available.
-        3. Otherwise, use the `default_plan` exposed through `GET /`.
+        Back-ends MUST resolve the billing plan in the following way:
+
+        * If a non-`null` value is given: Persist the `plan` that has been provided in the request.
+        * Otherwise:
+          1. Persist the `default_plan` exposed through `GET /me`, if available.
+          2. Persist the `default_plan` exposed through `GET /`, if available.
+          3. If a single plan is exposed by the back-end, persist it.
+          4. Otherwise, the back-end MUST throw a `BillingPlanMissing` error.
 
         The resolved plan MUST be persisted permanently, regardless of any 
         changes to the exposed billing plans in `GET /` in the future.
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -740,10 +740,20 @@ paths:
                         example: USD
                       default_plan:
                         type: string
-                        description: >-
-                          Name of the default plan to use when the user doesn't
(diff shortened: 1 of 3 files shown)


m-mohr:
@soxofaan Does this PR help to close the issues you raised in #425 and #426?

openapi.yaml
```diff
@@ -3492,12 +3502,12 @@ paths:
                     type: string
                     nullable: true
                     description: |-
-                      Name of the plan the user has subscribed to.
+                      Name of the plan the user has subscribed to, if any.
```
soxofaan, line 3505: "has subscribed to" -> "is currently subscribed to"

(it's not necessarily the same, e.g.  expiring plans)

openapi.yaml
```diff
@@ -3492,12 +3502,12 @@ paths:
...
 
-                      Overrides the default plan of the back-end, but back-ends
-                      MAY also allow overriding this plan for each individual
-                      processing request (e.g. job or service) with the
-                      corresponding `plan` property.
+                      Back-ends resolve billing plans in the following order:
+                      1. If provided, use the `plan` in the processing request.
+                      2. If provided, use the plan exposed here.
+                      3. Otherwise, use the `default_plan` exposed through `GET /`.
```
soxofaan, line 3510: I'm not sure it's useful to explain this priority scheme here. I think it complicates the description unnecessarily. It's discussed in a more appropriate place already. 

What could be useful to mention is whether a user can be subscribed to multiple plans (I think not, if I understand correctly). 
Another subtlety is whether there is a difference between a missing "default_plan" or a "null" default plan (e.g. backend does not support plans vs user has no plan)

openapi.yaml
```diff
@@ -4075,14 +4085,16 @@ components:
...
-        changes to the default billing plan in the future. If given, the default
-        plan specific to the user has higher priority than the default plan of
-        the back-end.
-
-        The list of plans and the default plan for the back-end can be retrieved
-        from `GET /`. The default plan for the user can be retrieved from `GET /me`.
+        Back-ends MUST resolve the billing plan in the following order:
+        1. Use the `plan` that has been provided in the request, if a non-null value is present.
+
+        If `null` has been provided for the `plan`:
```
soxofaan, line 4091: Isn't it strange to start a new paragraph before bullet `2`?

openapi.yaml
```diff
@@ -4075,14 +4085,16 @@ components:
...
+        2. Use the `default_plan` exposed through `GET /me`, if available.
+        3. Otherwise, use the `default_plan` exposed through `GET /`.
```
soxofaan, line 4093: What if it's still unresolved after `3`?

openapi.yaml
```diff
@@ -4075,14 +4085,16 @@ components:
...
+
+        The resolved plan MUST be persisted permanently, regardless of any
```
soxofaan, line 4095: > MUST be persisted permanently

That sounds quite strong and vague at the same time. I guess this is just about explicitly associating a plan to a processing request (e.g. exposed in metadata)? Or is it also about legal traceability?

openapi.yaml
```diff
@@ -740,10 +740,20 @@ paths:
...
+                          2. also did not provide a specific plan with the
+                             processing request.
+
+                          If a free plan is available at the back-end, it is 
+                          probably most useful to provide this as the back-end
+                          wide default plan and override it with paid plans through
+                          the user-specific default plan in `GET /me`.
+                          Otherwise, users that have not set up payment yet MAY
+                          receive an error for each processing requests where
+                          they did not provide a free plan specifically.
```
soxofaan, line 756: As  I noted elsewhere, but probably good to pin this note here:

I'm questioning if it's useful to define a default plan at the global level, if there is already a default plan at the user level. Conceptually that would only be useful if you want to associate a plan to anonymous requests, but what sense does that make?

I understand that removing the default plan from `/`  is technically a breaking change, but practically that's a non-issue as there is no back-end that already handles plans/billing for real.

soxofaan (commented):
some notes

openapi.yaml
m-mohr:
```suggestion
                      Name of the plan the user is currently subscribed to if any.
```

openapi.yaml
m-mohr: It's not really a new paragraph, but you need a newline to not have the text simply appended to point 1. But anyway, I just figured out that ReDoc doesn't render the list correctly, so will need to change something anyway.

openapi.yaml
m-mohr: If billing is supported, an error should be thrown. Added the BillingPlanMissing error.

openapi.yaml
m-mohr: Tried to clarify:

        The resolved plan MUST be persisted permanently so that the it is explicitly
        associated to a processing request and doesn't change when changes are made
        to the exposed billing plans in `GET /` in the future.

openapi.yaml
m-mohr: Fixed

openapi.yaml
m-mohr: * Agreed, removed the priority scheme.
* Clarified that only a single plan is supported.
* null: I think I can actually remove the nullable as it has the same meaning as "not present".

openapi.yaml
m-mohr: It might not be most useful, but it can simply be omitted for now. Are you aiming at deprecating the global one for 1.x?

openapi.yaml
soxofaan: Another thing is that "persist"  here is a bit tech-speak with a bit different meaning than the general verb "persist". I think the statement works just fine as 

> The resolved plan MUST be associated explicitly to ...

openapi.yaml
soxofaan: Indeed, a back-end can choose to not support it. 
The issue is mainly that it makes the spec/documentation confusing IMHO. If the global one is removed, you don't need the three step priority scheme, you just need one straightforward sentence.

m-mohr: Yes, but we can't simply remove it for now. Added it to the list of 2.0.0 clean-up tasks: #374 

Also, I could see us moving the whole billing part into the user details as also the currency or billing plans could be user specific.

openapi.yaml
m-mohr: We are in a developer resource here so persist should be fine especially as it's explained directly afterwards.

Issue #329, opened 2020-07-31
Logs: Add parameter for minimum severity


m-mohr:
It could be useful to filter logs returned to a client by a minimum severity. So for example, if I pass `severity=warning` to the back-end, it should only return warnings and errors, but not info or debug messages.

soxofaan:
Do you mean filtering when _retrieving_ the logs? This would imply that the backend always has to keep track of all log levels, which might not be desirable for large jobs (debug level might be very verbose).

Alternative (or additional) approach: user has to specify minimum log level (e.g `warning` by default) when submitting the job

m-mohr:
Yes, I was speaking about the retrieval. If the back-end doesn't provide some levels, that is fine. The contract doesn't enforce any level. Allowing the user to select what he wants to log when submitting the process is also a good idea.

m-mohr:
This idea came up again in an internal discussion where @clausmichele proposed:

> > Could we add the info about each partial result in the log if a debug flag has been activated? [...]

PR #431, opened 2021-11-30
Logs: Add data processing parameter for minimum severity #329

m-mohr:
Implements #329 as a parameter to set on the data processing request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /me`: New Relation types `alternate` and `related` for user-specific external pages. [#404](https://github.com/Open-EO/openeo-api/issues/404)
 - `GET /credentials/oidc`: Allow `authorization_code` and `urn:ietf:params:oauth:grant-type:device_code` (both without PKCE) as grants for `default_clients`. [#410](https://github.com/Open-EO/openeo-api/issues/410)
 - `GET /jobs/{job_id}/results`: Recommendation to add a link with relation type `canonical` which points to a signed URL with the same content as the response. [#397](https://github.com/Open-EO/openeo-api/issues/397)
-- `GET /jobs/{job_id}/logs`, `GET /services/{service_id}/logs` and `POST /result`: Added `log_levels` property in responses to indicate the log levels that are included. [#329](https://github.com/Open-EO/openeo-api/issues/329)
+- `GET /jobs/{job_id}/logs`, `GET /services/{service_id}/logs` and `POST /result`: Added `log_level` property in responses to reflect the minimum log level that has been chosen by the user. [#329](https://github.com/Open-EO/openeo-api/issues/329)
 - Added property `log_level` to secondary web service, batch job and synchronous processing endpoints to indicate the minimum severity level that should be stored for logs. [#329](https://github.com/Open-EO/openeo-api/issues/329)
 - Recommendation to add media types and titles to links for a better user experience.
 - Allow the relation type `canonical` to be used generally for (shared) resources (e.g. UDPs or batch jobs) without requiring Bearer authentication. [#405](https://github.com/Open-EO/openeo-api/issues/405)
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clarify that the relation type `version-history` should include `/.well-known/openeo` in the URL.
 - Clarify that clients should (re-)request capabilities and discovery endpoints with token if available and supported. [#416](https://github.com/Open-EO/openeo-api/issues/416)
 - `GET /`: Removed the superfluous default value for `currency`. [#423](https://github.com/Open-EO/openeo-api/issues/423)
+- Reflect that the `debug` process has been renamed to `inspect`.
 
 ## [1.1.0] - 2021-05-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `GET /jobs/{job_id}/results`: Recommendation to add a link with relation type `canonical` which points to a signed URL with the same content as the response. [#397](https://github.com/Open-EO/openeo-api/issues/397)
 - `GET /`: New Relation types: [#404](https://github.com/Open-EO/openeo-api/issues/404)
   - `create-form` to link to the registration page
   - `recovery-form` to link to the credentials recovery page.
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -5611,6 +5624,47 @@ components:
       example:
         - href: 'https://example.openeo.org/docs/errors/SampleError'
           rel: about
+    log_level:
+      description: |-
+        The severity level of the log entry.
+
+        The order of the levels is as follows (from low to high severity): `debug`, `info`, `warning`, `error`.
+        
+        The level `error` usually stops processing the data.
```
soxofaan, line 5633: I'm not sure if it's useful to claim that the "error" level corresponds with termination. There could still be retry attempts after an error situation. Some systems also include the level "critical" for errors that cause termination.

openapi.yaml
```diff
@@ -5611,6 +5624,47 @@ components:
...
+        The order of the levels is as follows (from low to high severity): `debug`, `info`, `warning`, `error`.
+
+        That means if `warning` is set, the back-end will only store log entries with the level `warning` and `error`.
+      type: string
+      enum:
+        - error
+        - warning
+        - info
+        - debug
+      default: debug
```
soxofaan, line 5654: Using "debug" level by default seems wasteful and bad for the default signal to noise ratio of the logs (especially because all logs are returned by default). I would at least use "info" by default.

openapi.yaml
```diff
@@ -5611,6 +5624,47 @@ components:
...
+        
+        The level `error` usually stops processing the data.
+      type: string
+      enum:
+        - error
+        - warning
+        - info
+        - debug
+      example: error
+    min_log_level_default:
```
soxofaan, line 5641: I'm a bit confused why there is this `_default` suffix here

openapi.yaml
```diff
@@ -6059,6 +6101,16 @@ components:
               - logs
               - links
             properties:
+              log_levels:
```
soxofaan, line 6104: Why a list of levels, instead of the minimum level to include?

Also: should it list the actual levels in the logs, or the possible levels?  For example, if the minimum log level is "warning" and there were no errors, should this list just `["warning"]`, or `["warning", "error"]`?

openapi.yaml
m-mohr: While it looks like an addition to this PR, this has been in the API definition before, see line 5925. It also just says that it "usually" stops, which I think it true in most cases.

openapi.yaml
m-mohr: That's nothing a user sees (it's a schema identifier) and is just to internally distinguish between schemas that have a default value (usual case) and those that don't have one (update case). This is aligned with several other schemas in the OpenAPI, which also have this suffix.

openapi.yaml
m-mohr: Yes, I was also thinking about this but I had two reasons for including debug (for now):
1. It was the behavior before this addition and it would not be ideal to change the behavior in-between minor versions.
2. The `debug` process is having the `debug` level assigned. Requiring a user to specifically enable the debug level just to see what he requested to log explicitly is confusing. (I'd be open to rename the `debug` process to `log` and assign it the `info` level though.)

openapi.yaml
m-mohr: This was meant guidance for clients, made available by back-ends in a best-effort manner. The array seemed more flexible as it could happen that there are only warnings and no errors reported and then the "minimum level" reporting as in job creation etc would be misleading.

Overall, it doesn't seem overly important and we may just remove it, too.

soxofaan: I think just stating the minimum level is good enough for clients, unless I'm missing something

m-mohr: Okay, updated the PR correspondingly.

openapi.yaml
soxofaan: Maybe my problem is that "The level `error` usually stops processing the data." sound like the logging _itself_ makes the processing stop.

Bit more verbose proposal:

> The level `error` usually corresponds with critical issues that unexpectedly terminate the normal data processing.

openapi.yaml
soxofaan: That's better, 
but still a bit strange that the default does not correspond with the recommendation.

Using "warning" by default would correspond to what I expect from a logging system, but maybe that will make user support harder.

Bit torn on this one

CHANGELOG.md
```diff
@@ -8,12 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - `GET /`: New Relation types: [#404](https://github.com/Open-EO/openeo-api/issues/404)
   - `create-form` to link to the registration page
   - `recovery-form` to link to the credentials recovery page.
 - `GET /me`: New Relation types `alternate` and `related` for user-specific external pages. [#404](https://github.com/Open-EO/openeo-api/issues/404)
 - `GET /credentials/oidc`: Allow `authorization_code` and `urn:ietf:params:oauth:grant-type:device_code` (both without PKCE) as grants for `default_clients`. [#410](https://github.com/Open-EO/openeo-api/issues/410)
+- `GET /jobs/{job_id}/results`: Recommendation to add a link with relation type `canonical` which points to a signed URL with the same content as the response. [#397](https://github.com/Open-EO/openeo-api/issues/397)
+- `GET /jobs/{job_id}/logs`, `GET /services/{service_id}/logs` and `POST /result`: Added `log_levels` property in responses to indicate the log levels that are included. [#329](https://github.com/Open-EO/openeo-api/issues/329)
```
soxofaan, line 17: still mentioning the old `log_levels` array here I think

openapi.yaml
```diff
@@ -6059,6 +6103,21 @@ components:
               - logs
               - links
             properties:
+              log_level:
+                description: |-
+                  The minimum severity level for log entries that the back-end returns.
+                  This property should reflect the `log_level` that has been specified by the user
```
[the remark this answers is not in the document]
m-mohr: I guess I could invent one (back-end only reports errors and nothing else), but MUST is also fine.

CHANGELOG.md
m-mohr: Indeed, fixed.

openapi.yaml
m-mohr: Yes, that's the difference in the schemas.

openapi.yaml
m-mohr: Yeah, I understand that, but also don't have a good solution. Maybe we can just try it out for now and evolve from there in 2.0. 🤷‍♀️

openapi.yaml
m-mohr: Ah, now I understand better what the issue is. That makes sense, will update.

soxofaan (approved):

Issue #430, opened 2021-11-29
Support partial/incomplete batch job results


soxofaan:
https://github.com/Open-EO/openeo-api/blob/f303d65a3291d4cd74dacc0e796803bb5d6fa03b/openapi.yaml#L3044-L3045

The API currently requires that a batch job is fully finished before results can be requested with `/jobs/{job_id}/results`.
In the context of very large batch jobs I think it could be useful to relax this and already allow listing of incomplete results (properly indicating that result is incomplete of course).

I'm not sure what is the best option: changing the behavior of the existing endpoint `/jobs/{job_id}/results`, or adding a parameter to enable this partial listing, or adding a new endpoint, or ...

re: https://github.com/openEOPlatform/architecture-docs/issues/12

soxofaan:
> In the context of very large batch jobs .. it could be useful to ... allow listing of incomplete results 

FYI: in the "large area" feature, we aim to split a very large job in smaller jobs at the level of the aggregator, and distribute this work as separate "sub" batch jobs on one or more back-ends.  It would be frustrating if  partial results would not be  accessible because one "sub" batch jobs fails or is stuck.

m-mohr:
Adding this to the existing endpoint sounds reasonable although it could be an issue that existing clients could start downloading and propagating partial results as "complete" as the semantic for providing partial results is not implemented. My first idea was using status code 206 instead of 200, but that doesn't fully fit. How would you envision this to work with things like `start_and_wait` in the clients?

m-mohr:
Yeah, I think a parameter works for me...

m-mohr:
A first draft is now available in PR #433

PR #433, opened 2021-11-30
Support partial/incomplete batch job results #430

m-mohr:
Implements suggestion from issue #430 as a parameter.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -3062,18 +3062,6 @@ paths:
   '/jobs/{job_id}/results':
     parameters:
       - $ref: '#/components/parameters/job_id'
-      - name: partial
-        description: >-
-          If set to `true`, the results endpoint returns incomplete results while
-          still running.
-
-          Enabling this parameter requires to indicate the status of the batch job
-          in the STAC metadata by setting the `openeo:status`.
-        in: query
-        allowEmptyValue: true
-        schema:
-          type: boolean
-          default: false
     get:
       summary: Download results for a batch job
       operationId: list-results
@@ -3130,6 +3118,19 @@ paths:
         - Batch Jobs
       security:
         - Bearer: []
+      parameters:
+        - name: partial
+          description: >-
+            If set to `true`, the results endpoint returns incomplete results while
+            still running.
+
+            Enabling this parameter requires to indicate the status of the batch job
+            in the STAC metadata by setting the `openeo:status`.
+          in: query
+          allowEmptyValue: true
+          schema:
+            type: boolean
+            default: false
       responses:
         '200':
           description: >-
@@ -3143,7 +3144,8 @@ paths:
                 to `0` if the requester is the owner of the job and still has
                 free downloads included in his processing charges estimated by
                 `GET /jobs/{job_id}/estimate`.
-                 If a requester other than the owner is requesting the data of a shared job this header indicates the costs for the requester.
+                If a requester other than the owner is requesting the data of a
+                shared job this header indicates the costs for the requester.
               schema:
                 $ref: '#/components/schemas/money'
           content:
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -3049,13 +3049,25 @@ paths:
...
     get:
-      summary: Download results for a completed batch job
+      summary: Download results for a batch job
       operationId: list-results
       description: |-
-        Lists signed URLs pointing to the processed files after the batch job
+        Lists signed URLs pointing to the processed files, usually after the batch job
         has finished. Back-ends may also point to intermediate results after the
-        job has stopped due to an error.
+        job has stopped due to an error or if the `partial` parameter has been set.
```
soxofaan, line 3070: I wasn't aware of this "intermediate results after the job has stopped due to an error" feature

I guess it also makes sense to enforce usage the "incomplete_result" field for that case too?


Related: to me, "finished" somewhat implies "finished successfully", but the description seems to also include "failed jobs", is that correct? The new `partial` option could make the interface less error prone:
- partial=false:  only complete results from successfully finished jobs are returned, all other cases give JobNotFinished error
- partial=true: incomplete results results from failed jobs are returned

openapi.yaml
```diff
@@ -3713,6 +3728,10 @@ components:
               minItems: 1
               items:
                 type: string
+    incomplete_result:
+      type: boolean
+      default: false
+      description: MUST be `true` when the batch job result has not finished yet and as such is potentially incomplete.
```
soxofaan, line 3734: See above: "finished" -> "finished successfully"

openapi.yaml
m-mohr: > I guess it also makes sense to enforce usage the "incomplete_result" field for that case too?

Not sure, but at least the result should indicate the state. Maybe we should just reflect the general job status in the result (i.e. the "status" property from the job metadata), e.g. as `openeo:status` instead of using the `openeo:incomplete` flag.

> "finished" somewhat implies "finished successfully"

I guess "finished" was meant to correspond to the openEO status "finished", which would be successfully finished indeed. I'm not sure whether this is used consistently in the API though. Will need to check.


Is it okay to change the API behavior here? Strictly this is breaking, but who would be relying on "errored" results being returned anyway? :-D So I think it should be fine to change anyway.

openapi.yaml
m-mohr: Will check the wording and/or make clear that this corresponds to the job status enum.

openapi.yaml
soxofaan: > Not sure, but at least the result should indicate the state.

Indeed I was also thinking along those lines: use a status enum field (e.g. finished/error/running, ..) instead of `openeo:incomplete` boolean.

m-mohr: Replaced openeo:incomplete with openeo:status.

openapi.yaml
m-mohr: This terminology is not present anymore after switching to `openeo:status`.

openapi.yaml
```diff
@@ -3726,6 +3741,19 @@ components:
...
+      enum:
+        - running
+        - canceled
+        - finished
+        - error
+      description: |-
+        The status of a batch job.
+
+        This field is REQUIRED if the `partial` parameter is given.
+        This field is strongly RECOMMENDED if the job has errored.
```
soxofaan, line 3755: Why this recommendation note for errored jobs? 
You can only get a response with `partial` on anyway, otherwise you get `JobNotFinished` (line 3127 above)

soxofaan (approved):
apart from a minor note about failed jobs, this looks ready to me

openapi.yaml
soxofaan: Also: I'm not sure "errored" is proper English (at least my spell checker complains)

m-mohr: > Why this recommendation note for errored jobs? You can only get a response with `partial` on anyway, otherwise you get `JobNotFinished` (line 3127 above)

Partial results can be exposed for jobs that ~errored~ stopped due to an error. In this case ideally, I'd require the status, too. But to not introduce a breaking change, it's just recommended for now.

>  I'm not sure "errored" is proper English

Replaced it with something else.

Issue #396, opened 2021-05-26
Explicitly specify metadata filters for collections


m-mohr:
Originates from https://github.com/Open-EO/openeo-python-client/issues/141#issuecomment-847340175

Initially the thought was that all (scalar) metadata properties could be queried. This seems not to be true and thus it's not really clear to users what they can query on (despite the limited implementations right now). It would probably good to add something that mentions what users can query on. 

In STAC API this would be `/queryables`, but it may be enough to just add a list with properties to the collections, e.g. `"openeo:properties": ["eo:cloud_cover", "proj:epsg"]` for example.

m-mohr:
It's actually possible to expose this via the process description, although it is pretty verbose:

```json
{
  "type": "object",
  "subtype": "metadata-filter",
  "title": "Filters",
  "description": "A list of filters to check against. Specify key-value-pairs with the key being the name of the metadata property name and the value being a process evaluated against the metadata values.",
  "properties": {
    "eo:cloud_cover": {
      "type": "object",
      "subtype": "process-graph",
      "parameters": [
        {
          "name": "value",
          "description": "The cloud cover value to be checked against.",
          "schema": {
            "type": [
              "number",
              "null"
            ],
            "minimum": 0,
            "maximum": 100
          }
        }
      ]
    },
    "platform": {
      "type": "object",
      "subtype": "process-graph",
      "parameters": [
        {
          "name": "value",
          "description": "The property value to be checked against.",
          "schema": {
            "type": [
              "string",
              "null"
            ]
          }
        }
      ]
    }
  }
}
```

m-mohr:
Although it would likely also be interesting to expose this per collection...

sophieherrmann:
I think it would make a lot more sense to expose this per collections. Thinking about radar and optical data the two will provide very different queryable properties. Depending on the specific datasets it could even be that on radar collection provides a specific queryable and another radar collection does not.

I do like the idea of just adding an additional key which lists non standard queryable properties.

m-mohr:
We may want to consider collection-specific queryables instead of `openeo:property_filters`. Came up in https://github.com/Open-EO/openeo-api/pull/447#issuecomment-1206476270

m-mohr:
Removed `openeo:property_filters` again. Instead add queryables once working on #469 and the underlying specs are stable (OGC API and STAC API).

m-mohr:
I'll prepare a PR based on the STAC API - Filter extension.

[...]

PR #434, opened 2021-11-30
Explicitly specify metadata filters for collections #396

m-mohr:
Adds a STAC metadata field to implement #396

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4353,15 +4353,15 @@ components:
         'openeo:property_filters':
           description: |-
             Gives a list of properties you can filter on when loading data from the collection.
-            Returning an empty array means that no filters are available.
+            Returning an empty array means that no property-based filters are available.
 
-            This is a guide for users to get an overview over available filters.
+            This is a guide for users to get an overview of available properties to filter on.
             More metadata about the properties can be found in `summaries`.
 
             If this property is not present, it means that the collections does 
-            not provide any collection-specific details about available filters. 
+            not provide any collection-specific details about available property-based filters. 
             It's then up to other means of documentation to inform the user,
-            e.g. there could be documentation about filters that work across collections.
+            e.g. there could be documentation about property filters that work across collections.
           type: array
           items:
             type: string
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4353,9 +4353,15 @@ components:
         'openeo:property_filters':
           description: |-
             Gives a list of properties you can filter on when loading data from the collection.
+            Returning an empty array means that no filters are available.
+
+            This is a guide for users to get an overview over available filters.
             More metadata about the properties can be found in `summaries`.
 
-            Returning an empty array means that no filters are available.
+            If this property is not present, it means that the collections does 
+            not provide any collection-specific details about available filters. 
+            It's then up to other means of documentation to inform the user,
+            e.g. there could be documentation about filters that work across collections.
           type: array
           items:
             type: string
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1445,6 +1445,10 @@ paths:
                       - B10
                       - B11
                       - B12
+                'openeo:property_filters':
+                  - platform
+                  - eo:cloud_cover
+                  - sat:orbit_state
                 summaries:
                   'constellation':
                     - Sentinel-2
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -4346,6 +4350,15 @@ components:
             x-additionalPropertiesName: Dimension Name
             allOf:
               - $ref: '#/components/schemas/dimension'
+        'openeo:property_filters':
+          description: |-
+            Gives a list of properties you can filter on when loading data from the collection.
+            More metadata about the properties can be found in `summaries`.
+
+            Returning an empty array means that no filters are available.
+          type: array
```
soxofaan, line 4359: minor question: should there be a default value specified (empty array)?

m-mohr: No, if the default is an empty array then that would mean that by default no data would be returned at all.
The default right now is "not given" => potentially all levels could be included, which was the case before, too.
If "not given" is too implicit we could allow null as a data type and default to that, which means no information available, expect any level to be present. This field anyway is just an indication for clients.

soxofaan: thanks, what remains bit unclear from current docs:


- "not given": all properties can be filtered on (backend is expected to raise error when property filter is invalid in some way)?
- side note: is it (useful to make it) possible to make this "everything is allowed, as long as backend doesn't complain" explicit, e.g. by specifying "null"?
- if a list is specified: are only these properties allowed to filter on? Using non-whitelisted properties should raise an error (e.g. client side) ? Or is just guidance for users and are other property filters still allowed?

m-mohr: Not giving it doesn't necessarily mean everything is possible. It just is the backward-compatible way of saying: We don't expose it, good luck with figuring it out yourself. ;-)

"Everything is possible" can just be encoded by listing all values explicitly, which I think is the most explicit way. Using null for it is rather confusing from a user POV as this would then put the work onto the user to figure out what is available. And yes, this is meant to be guidance for users (and clients).

Raising errors is part of the processes and would need to be documented there, so we don't imply the best approach for the processes here.

openapi.yaml
```diff
@@ -4353,9 +4353,15 @@ components:
         'openeo:property_filters':
           description: |-
             Gives a list of properties you can filter on when loading data from the collection.
+            Returning an empty array means that no filters are available.
```
soxofaan, line 4356: "... no filters ... " - >"no property based filters"

openapi.yaml
```diff
@@ -4353,9 +4353,15 @@ components:
...
+
+            This is a guide for users to get an overview over available filters.
```
soxofaan, line 4358: "available filters" ->  "available properties to filter on"

openapi.yaml
```diff
@@ -4353,9 +4353,15 @@ components:
...
             More metadata about the properties can be found in `summaries`.
 
-            Returning an empty array means that no filters are available.
+            If this property is not present, it means that the collections does 
+            not provide any collection-specific details about available filters.
```
soxofaan, line 4362: "available filters" -> "available property based filters"

openapi.yaml
```diff
@@ -4353,9 +4353,15 @@ components:
...
+            Returning an empty array means that no filters are available.
+
+            This is a guide for users to get an overview over available filters.
             More metadata about the properties can be found in `summaries`.
 
-            Returning an empty array means that no filters are available.
+            If this property is not present, it means that the collections does 
+            not provide any collection-specific details about available filters. 
+            It's then up to other means of documentation to inform the user,
+            e.g. there could be documentation about filters that work across collections.
```
soxofaan, line 4364: "filters" -> "property filters"

soxofaan (approved):
just some minor suggestions

openapi.yaml
m-mohr:
```suggestion
            Returning an empty array means that no property-based filters are available.
```

openapi.yaml
m-mohr:
```suggestion
            This is a guide for users to get an overview of available properties to filter on.
```

openapi.yaml
m-mohr:
```suggestion
            not provide any collection-specific details about available property-based filters. 
```

openapi.yaml
m-mohr:
```suggestion
            e.g. there could be documentation about property filters that work across collections.
```

PR #439, opened 2022-02-17
Update broken links to file format GDAL codes

soxofaan:
http://www.gdal.org/formats_list.html  and https://gdal.org/ogr_formats.html  give 404


soxofaan:
If I understand correctly these pages were removed here: https://github.com/OSGeo/gdal/pull/1562

PR #441, opened 2022-02-23
Vector data cubes

m-mohr:
Minor additions to the API only, Mostly reflecting the changes in https://github.com/stac-extensions/datacube/pull/10 and the addition to the JSON Schema for the new `datacube` subtype.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -5875,12 +5875,17 @@ components:
 
             The order in the array corresponds to the parameter order to
             be used in clients that don't support named parameters.
+
+            This property only applies if the `subtype` is set to `process-graph`.
           items:
             $ref: '#/components/schemas/parameter'
         returns:
           type: object
           title: Process Graph Return Value
-          description: Description of the data that is returned by the child process graph.
+          description: |-
+            Description of the data that is returned by the child process graph.
+
+            This property only applies if the `subtype` is set to `process-graph`.
           required:
             - schema
           properties:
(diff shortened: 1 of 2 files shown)


m-mohr:
This is ready for final review.

openapi.yaml
```diff
@@ -1375,7 +1376,7 @@ paths:
               example:
                 stac_version: 1.0.0
                 stac_extensions:
-                  - https://stac-extensions.github.io/datacube/v2.0.0/schema.json
+                  - https://stac-extensions.github.io/datacube/v2.2.0/schema.json
```
dthiex, line 1379: This link does not work yet.

m-mohr: Yes, the extension is also waiting for review and release.

openapi.yaml
```diff
@@ -4820,6 +4855,37 @@ components:
...
+      title: Geometry type
+      type: string
+      enum:
+        - Point
+        - MultiPoint
+        - LineString
+        - MultiLineString
+        - Polygon
+        - MultiPolygon
+        - GeometryCollection
```
soxofaan, line 4868: Shouldn't there be  a note here (like in other places) that GeometryCollection should be avoided?

soxofaan (approved):
approve (with minor remark)

openapi.yaml
m-mohr: No, I don't think it makes a lot of sense here. This is backend-oriented documentation. Back-ends can't really "avoid" GeometryCollections, they either support it or not. The data that may consist of GeometryCollections is provided in a different place (processes) by a different audience (end-users) and for this case it's documented in the processes.

m-mohr:
This is also ready for final review. There are some issues with the relevant tooling (redoc), so I can't express everything 1:1 in the API spec. The name for the vector dimension is now "geometries" (plural) in accordance with "bands", okay?

@soxofaan @mkadunc @aljacob @clausmichele @jdries @dthiex @LukeWeidenwalker @pierocampa

soxofaan:
> The name for the vector dimension is now "geometries" (plural) in accordance with "bands", okay?

I don't think that changing to plural "geometries" is really necessary. Apart from "bands", all dimension types and names are/look singular:  "spatial", "temporal", "x", "y", "other".

As another data point here: geopandas uses singular "geometry" as column type and name in data frames, e.g. see https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.html#geopandas.GeoDataFrame

pierocampa (approved):

aljacob (approved):

LukeWeidenwalker (approved):

m-mohr:
> I don't think that changing to plural "geometries" is really necessary.

Are you saying 
>            * `geometries` for dimensions of type `geometries`

should change to
>           * `geometry` for dimensions of type `geometries`

?

soxofaan:
I guess so. 

If I understand correctly, the dimension **type** has already been settled on plural "geometries" in merged PR https://github.com/stac-extensions/datacube/pull/10
so in this PR #441 there is only room to change the dimension **name** to singular "geometry", which would indeed not look very consistent in the recommendation
```
* name `geometry` for dimensions of type `geometries`
```

Sidenote: I noticed that @aljacob made related comment at https://github.com/Open-EO/openeo-processes/pull/382#pullrequestreview-1261722159

m-mohr:
In principle we can still change everything as we just merged the PRs yet, but we have not released anything yet (including the STAC extension). Anyway, I'm merging this now and will then create a separate issue for this. Makes it easier to work on further improvements.

PR #444, opened 2022-03-24
Clarified ambiguous batch job status changes

### LINKED ISSUE #442: response on starting a job that is already queued/running/finished
`POST /jobs/{job_id}/results` is mean to get a `created` job into `queued`/`running` state, and response with `202 The creation of the resource has been queued successfully.`

But its unclear what the response should be when doing `POST /jobs/{job_id}/results` on a job that is already queued/running/finished/error.

> This endpoint has no effect if the job status is already in status `queued` or `running`. 

Doing nothing at back-end side is fine, but what should the response be? Also  `202 The creation of the resource has been queued successfully.`? 

And what in case of a job that is already in status `finished` or `error`?
While also  "having no effect" is still fine I guess, it's a bit weird to respond with  `202 The creation of the resource has been queued successfully.`. Returning with 4XX/5XX also doesn't really fit, I'd think.

(related to #436)

m-mohr:
Fixes #442 (partially?) and also fixes issues reported by @danielFlemstrom via e-mail.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2882,13 +2882,14 @@ paths:
     patch:
       summary: Modify a batch job
       operationId: update-job
-      description: >-
+      description: |-
         Modifies an existing job at the back-end, but maintains the identifier.
         Changes can be grouped in a single request.
 
+        The job status does not change.
 
         Jobs can only be modified when the job is not queued or running.
-        Otherwise requests to this endpoint MUST be rejected with openEO error
+        Otherwise, requests to this endpoint MUST be rejected with openEO error
         `JobLocked`.
       tags:
         - Data Processing
@@ -3268,18 +3269,21 @@ paths:
         The result will be stored in the format specified in the process.
         To specify the format use a process such as `save_result`.
 
-        This endpoint has no effect if the job status is already `queued` or
-        `running`. In particular, it doesn't restart a running job. Processing
-        MUST be canceled before to restart it.
-
         The job status is set to `queued`, if processing doesn't start
-        instantly.
+        instantly. The same applies if the job status is `canceled`, `finished`,
+        or `error`, which restarts the job and discards previous results.
+        Clients SHOULD warn users and ask for confirmation if results may get
+        discarded.
 
         * Once the processing starts the status is set to `running`.
         * Once the data is available to download the status is set to
         `finished`.
         * Whenever an error occurs during processing, the status MUST be set to
         `error`.
+
+        This endpoint has no effect if the job status is already `queued` or
+        `running`. In particular, it doesn't restart a running job. To restart
+        a queued or running job, processing MUST have been canceled before.
 
         Back-ends SHOULD reject queueing jobs with openEO error `PaymentRequired`,
         if the back-end is able to detect that the budget is too low to fully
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -2882,13 +2882,14 @@ paths:
...
       summary: Modify a batch job
       operationId: update-job
-      description: >-
+      description: |-
         Modifies an existing job at the back-end, but maintains the identifier.
         Changes can be grouped in a single request.
 
+        The job status does not change.
 
         Jobs can only be modified when the job is not queued or running.
```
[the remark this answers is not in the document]
soxofaan: And on a more general note: why can jobs only be changed when not queued/running? This is only about changing the metadata of a job, it should not interfere with the processing itself, right?

openapi.yaml
```diff
@@ -3268,19 +3269,22 @@ paths:
...
 
-        This endpoint has no effect if the job status is already `queued` or
-        `running`. In particular, it doesn't restart a running job. Processing
-        MUST be canceled before to restart it.
-
         The job status is set to `queued`, if processing doesn't start
-        instantly.
+        instantly. The same applies if the job status is `canceled`, `finished`,
+        or `error`, which restarts the job and discards previous results.
+        Clients SHOULD warn users and ask for confirmation if results may get
```
soxofaan, line 3275: I'm not a fan of moving this responsibility to the client (library). The client is then forced to get the status before sending the start instruction and there must be a way to interactively ask the user for confirmation (which is not straightforward in scripting/headless contexts). Due to latency there is also risk on race conditions.

As mentioned elsewhere I would propose that a user first has to actively cancel a finished job before it can be restarted (and a "start" on a finished job does nothing or raises error)

m-mohr: It's not really doable in a compliant way, e.g. right now the DELETE endpoint for .../results says it has no effect for states other than queued or running. In the future we could make this endpoint a general cancel + delete endpoint, indeed. So I'd propose to open an issue for a 2.x version?

openapi.yaml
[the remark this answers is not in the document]
m-mohr: You can also change the process graph and that is likely not a good idea while running. Similarly it's not a good idea to change the plan while runnnig. Also, the title etc are likely to get stored to the metadata, which may lead to strange behavior?

```suggestion
        Jobs can only be modified when the job is not queued and not running.
```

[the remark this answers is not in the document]
soxofaan: > change the process graph ... Similarly, it's not a good idea to change the plan while running etc.

That would indeed be a bad idea for the user in the sense that they make things very confusing.  For the back-end it's probably less of a problem as it can use the process graph or plan at the time of starting the job.

Nevertheless, if this is a problem, then I think it's  also a bad idea to change these things even when the job is finished or failed.  So the constraint should be: jobs can only be modified when created or canceled

openapi.yaml
soxofaan: ok I understand that the feature "cancel before restart" requires v2-style changes.

But here we are defining the behavior what happens when sending "start" action when status is finished/canceled/error.

This PR explicitly adds destructive behavior for the "finished" case, while it previously was not destructive if you extrapolate from:
> This endpoint has no effect if the job status is already `queued` or
        `running`. In particular, it doesn't restart a running job. Processing
        MUST be canceled before to restart it.

I know it doesn't mention the "finished", but it sets a sentiment to handle batch job results defensively: starting a job should not break/remove what was already achieved.

I'd prefer to make this defensive behavior more explicit (e.g. no effect, or throw error) instead of the destructive handling for status "finished".

I'm fine with handling start on canceled/error status as proposed in PR.

m-mohr: I don't understand where this PR adds destructive behavior? This is simply meant to clarify existing behavior. 'finished' wasn't mentioned intentionally and as such, it was destructive before this PR already.

soxofaan: > 'finished' wasn't mentioned intentionally and as such, it was destructive before this PR already.

I don't completely agree: the behavior was undocumented, so it was up to the back-end implementation what actual behavior you would get.  For example with the VITO back-end, since https://github.com/Open-EO/openeo-python-driver/commit/cf1a2036d6b0e8d2b5ad99fb98bacae34ec60c41  (Feb 2022), a start on a finished job does nothing. Before that change, it was indeed destructive (with chance of having some left-over results from the first run).  I can imagine that other back-ends had different behavior (e.g. 500 Internal error for some reason)

PR #447, opened 2022-05-04
Commercial data API

Ardweaden:
This PR adds documentation for the commercial data API, which should provide the ability to purchase commercial data. It is mostly based on Sentinel Hub's API, but it's greatly simplified.

This is most certainly not the final version, as there are quite a few things that should be discussed.

### Pricing
General information about pricing is only provided in a human-readable way. We could attempt to standardise it, but it might not be so trivial. Actual costs of a specific order are returned in the values of currency upon creating the order.

### Search and order parameters
These are again not provided in a computer-readable way as there can be very many different options.
However, it would make sense to generalise spatial extent and temporal extent, as they will most likely appear on most collections - however I'm not sure they will be always usable, that's why this proposal does not implement that.

### Accessing commercial data
This proposal relies on the setup that VITO already employs, of having information about a "general collection" available at `/collections`, and then a specific identifier needs to be used in `load_collection` in order to access the actual purchased data. That's why the orders in this proposal include `byoc_collection_id`, which I think is not really appropriate.
The referenced issue is discussed here: https://github.com/openEOPlatform/architecture-docs/issues/4#issuecomment-983451983


m-mohr:
I'm not actually sure that this should be merged to the openEO API as it seems very specific to Platform and has not had a huge discussion across back-end implementations. As you mentioned, not everything is machine-readable yet and as such it is not really an interoperable API that can be easily be used by our clients etc. Also, it seems not fully integrated with the existing approaches (e.g. integrate with /collections and STAC better). Would it make more sense to have this separate as an extension proposal for openEO or as a separate proprietary API for openEO Platform?

Ardweaden:
> I'm not actually sure that this should be merged to the openEO API as it seems very specific to Platform and has not had a huge discussion across back-end implementations. As you mentioned, not everything is machine-readable yet and as such it is not really an interoperable API that can be easily be used by our clients etc. Also, it seems not fully integrated with the existing approaches (e.g. integrate with /collections and STAC better). Would it make more sense to have this separate as an extension proposal for openEO or as a separate proprietary API for openEO Platform?

@mkadunc @dthiex 

> Would it make more sense to have this separate as an extension proposal for openEO

This would make sense. Are there any other existing openEO extensions?

m-mohr:
There's the Federation Extension draft, which you can look at in the draft branch. The main issue with extensions is that you can't extend OpenAPI easily and as such all the OpenAPI extensions are in code block in the Readme. 

My thought here is that if you want to use this exact API as proposed, you better do a separate release as a community extension.
Otherwise, I'd surely go over it and propose some changes that may change the API in a way that you need to do changes. ;-)
For example, I'm really wondering why there's everything grouped under /commercial_data while I think you could also merge it into some of the existing endpoints (e.g. /collections). And then just mark collections as "commercial" and then use the STAC order extension (once available). The search endpoint looks like it could just be a STAC API search endpoint? And then there's only /orders left that can be added as new endpoints. That are just the high-level thoughts, I haven't really digged into the details yet.

openapi.yaml
```diff
@@ -3497,8 +3500,739 @@ paths:
...
+                        description: General information about pricing for ordering products.
+                        properties:
+                          description:
+                            type: string
+                            description: Human-readable description of pricing.
+                          links:
+                            type: array
+                            description: |-
+                              Additional information about the pricing options.
+                  search_parameters:
```
m-mohr, line 3573: This could be handled via OGC's /queryables instead of a new interface.

openapi.yaml
```diff
@@ -3497,8 +3500,739 @@ paths:
...
+                  - id
+                  - pricing
+                  - search_parameters
+                  - order_parameters
+                properties:
+                  id:
+                    type: string
+                    description: A unique identifier for the collection, which MUST match the specified pattern.
+                    pattern: '^[\w\-\.~\/]+$'
+                  provider:
```
m-mohr, line 3540: This should probably be aligned with STACs `providers` field.

openapi.yaml
```diff
@@ -3497,8 +3500,739 @@ paths:
           $ref: '#/components/responses/client_error_auth'
         5XX:
           $ref: '#/components/responses/server_error'
+  '/commercial_data/collections/{collection_id}':
```
m-mohr, line 3503: Can we simply merge this with /collections/{collection_id} ?

openapi.yaml
```diff
@@ -3497,8 +3500,739 @@ paths:
...
+                    required: true
+                    links:
+                    - title: Bounds specification
+                    - rel: related
+                    - href: https://api.docs.com/tpdi/airbus#bounds
+        4XX:
+          $ref: '#/components/responses/client_error_auth'
+        5XX:
+          $ref: '#/components/responses/server_error'
+  /commercial_data/search:
```
m-mohr, line 3658: Can this be aligned with STAC's /search endpoint?

openapi.yaml
```diff
@@ -3497,8 +3500,739 @@ paths:
...
+                    - 3
+                    - 15
+                    - 4
+                    - 16
+                  time_range: 
+                    - "01-01-2022"
+                    - "01-02-2022"
+                  processing_level: SENSOR
+                  max_snow_coverage: 0
+  /commercial_data/orders:
```
m-mohr, line 3766: Any thoughts on aligning the general structure with https://github.com/stac-extensions/order ?

Ardweaden:
> My thought here is that if you want to use this exact API as proposed (as you likely have it implemented already), you better do a separate release as a community extension.

We haven't implemented it yet so that's not an issue.

> Otherwise, I'd surely go over it and propose some changes that may change the API in a way that you need to do changes. ;-)
> For example, I'm really wondering why there's everything grouped under /commercial_data while I think you could also merge it into some of the existing endpoints (e.g. /collections). And then just mark collections as "commercial" and then use the STAC order extension (once available). The search endpoint looks like it could just be a STAC API search endpoint? And then there's only /orders left that can be added as new endpoints. That are just the high-level thoughts, I haven't really digged into the details yet.

Thanks for great suggestions, we'll look into it and try to align the proposal to match them,

m-mohr:
Ah, good to know. Then I'll also invest some time into a deeper review soon.

openapi.yaml
Ardweaden: Good idea, done so.

openapi.yaml
Ardweaden: Indeed, thanks for that. And it works nicely with STAC /search.

However, the full OGC Queryables require 2 endpoints:
* `/queryables` which are available for all collections
* `{collectionId}/queryables` only for a specific one

I don't see a great benefit in having both, so I only added the second one. Do you think we should follow the spec closer and implement it fully?

I guess we could use the same for ordering parameters, what do you think? In our case they are usually the same as search parameters.

openapi.yaml
Ardweaden: It's ok, I used it but it's very limited in terms of information that it handles. Of course we could always extend it, but it's not quite clear yet what the best way of accessing commercial data would be, as mentioned in the description of this pull request (currently the solution works best for SH).

m-mohr:
Hey @Ardweaden, we were talking about this PR recently and I'm about to review it, but I was wondering whether it would be good to set up a call so that we can discuss the overall idea better instead of wrangling through 1000 lines of OpenAPI directly? I can't see your e-mail address here otherwise I'd have directly sent an e-mail.

openapi.yaml
m-mohr: I think we can loosely build on top of it.

m-mohr:
@Ardweaden fyi: here's the only existing extension, which you can get inspiration from: https://github.com/Open-EO/openeo-api/tree/draft/extensions/federation

Ardweaden:
> @Ardweaden fyi: here's the only existing extension, which you can get inspiration from: https://github.com/Open-EO/openeo-api/tree/draft/extensions/federation

@m-mohr I pushed the first version of README. I decided to revert the previous commits and therefore in the branch for now, so that we can refer to the changes made there if it comes in handy.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+          "UINT16",
+          "UINT8"
+        ]
+      },
+    ],
+  "pricing": {
+    "searching": null,
+    "ordering": {
+      "description": "Minimum area per order is 0.25 km2. The price is calculated based on a 6-months sliding window.",
+      "links": [
```
m-mohr, line 69: I think I'd propose to define relation types for this kind of links and instead add them to the links structure of the Collection.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+        }
+      ]
+    }
+  }
+}
+```
+
+### Filtering parameters discovery
+
+Searching for products can support refining the search by filtering with general or collection-specific attributes. Backends should implement a top level `/queryables` endpoint for attributes available for all collections, and collection-specific attributes should be provided at `/collections/{collection-id}/queryables` according to [OGC Queryables specification](https://portal.ogc.org/files/96288#filter-queryables).
```
m-mohr, line 83:
```suggestion
Searching for products can support refining the search by filtering with general or collection-specific attributes. Backends should implement a top-level `/queryables` endpoint for attributes available for all collections, and collection-specific attributes should be provided at `/collections/{collection-id}/queryables` according to the [OGC Queryables specification](https://portal.ogc.org/files/96288#filter-queryables).
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+The client can then create an order for the desired products. Because of the financial cost of purchasing the data, a separate endpoint for confirming the execution of that specific order should be implemented.
+
+When the order is completed, the user should be able to access the data using the `collection-id` at the temporal and spatial location of the purchased products.
+
+### Collection discovery
+
+A backend should add general information about a commercial data collection to the `/collections` and `/collections/{collection-id}` endpoints, the same as with freely available collections. Only the metadata about the entire dataset needs to be provided, not about the specific data products that a user has already purchased. 
+
+Commercial data collections are distinguished from freely available collections by including `"order:status": "orderable"` as specified in [STAC Order specification](https://github.com/stac-extensions/order).
```
m-mohr, line 25:
```suggestion
Commercial data collections are distinguished from freely available collections by including `"order:status": "orderable"` as specified in the [STAC Order specification](https://github.com/stac-extensions/order).
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+Commercial data collections can include `order_parameters` field, if ordering supports additional parameters that specify how the products should be delivered.
```
m-mohr, line 27:
```suggestion
Commercial data collections can include an `order_parameters` field if ordering supports additional parameters that specify how the products should be delivered.
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+}
+```
+
+### Filtering parameters discovery
+
+Searching for products can support refining the search by filtering with general or collection-specific attributes. Backends should implement a top level `/queryables` endpoint for attributes available for all collections, and collection-specific attributes should be provided at `/collections/{collection-id}/queryables` according to [OGC Queryables specification](https://portal.ogc.org/files/96288#filter-queryables).
+
+#### Example
+
+Example response from `GET collections/PLEIADES/queryables`:
```
m-mohr, line 87:
```suggestion
Example response from `GET /collections/PLEIADES/queryables`:
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+{
+   "$schema":"http://json-schema.org/draft-07/schema#",
+   "$id":"http://example.com/commercial_data/PLEIADES/queryables",
+   "type":"object",
+   "title":"Queryables of Airbus Pleiades collection.",
+   "description":"Available properties for CQL filtering of products.",
+   "properties":{
+      "processing_level":{
+         "title":"Processing level",
+         "description":"When searching, you will receive results from the full catalog as well as the Living Library, which holds images that have cloud cover under 30% and Incidence angle under 40°. If you want to search only Living Library results, you will need to filter using processingLevel. This value could be equal to SENSOR (images which meet Living Library criteria) and ALBUM (images that do not meeting Living Library criteria in terms of incidence angle and cloud cover).'",
```
m-mohr, line 98: To get a more concise example, maybe shorten the text to a sentence?

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+         "maximum":90,
+         "default":90
+      }
+   }
+}
+```
+
+### Searching available products
+
+Backends should implement top-level `GET /search` endpoint as specified in [STAC Item Search API specification](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/item-search). This should include the [Filter Extension](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/fragments/filter), which enables filtering the available products by attributes specified in `/queryables` and `collections/{collection-id}/queryables`.
```
m-mohr, line 126:
```suggestion
Backends should implement the top-level `GET /search` endpoint as specified in the [STAC Item Search API specification](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/item-search). This should include the [Filter Extension](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/fragments/filter), which enables filtering the available products by attributes specified in `GET /queryables` and `GET /collections/{collection-id}/queryables`. 
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+         "default":90
+      }
+   }
+}
+```
+
+### Searching available products
+
+Backends should implement top-level `GET /search` endpoint as specified in [STAC Item Search API specification](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/item-search). This should include the [Filter Extension](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/fragments/filter), which enables filtering the available products by attributes specified in `/queryables` and `collections/{collection-id}/queryables`. 
+The endpoint should return a list of  `STAC Item`s that match the criteria.
```
m-mohr, line 127:
```suggestion
The endpoint returns a list of STAC Items that match the criteria.
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+    3,
+    15,
+    4,
+    16
+  ],
+  "datetime": [
+    "01-01-2022",
+    "01-02-2022"
+  ],
+  "collectionsArray": [
```
m-mohr, line 145: This parameter is called collections in STAC.

```suggestion
  "collections": [
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+    ]
+  }
+}
+```
+
+### Ordering products
+
+Backends should implement the following endpoints:
+
+- `GET /orders`: Get a list of all created orders
```
m-mohr, line 178: I'm still wondering whether all endpoints should generally live under `/collections/{collection_id}/orders` with an additional `GET /orders` that lists all orders with limited metadata. Not sure yet, but posting it here for discussion.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+      type: object
+      description: Key-value pairs of available `order_parameters` as listed at `GET /commercial_data/collections/{collection_id}` for filtering available products.
+      additionalProperties:
+        x-additionalPropertiesName: Order Parameter Name
+        description: Value of the order parameter to be used in the request.
+```
+
+#### `GET /orders/{order-id}`
+
+Get full metadata of the order. The item should follow the [STAC Order Extension](https://github.com/stac-extensions/order), but extended with the spatial and temporal extent information and other metadata about the products.
```
m-mohr, line 284: I'm wondering whether this should actually return a STAC Collection?

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+## Overview of the workflow
+
+All the available datasets provided by a backend are listed on the `/collections` endpoint. The collections are normally freely accessible. This extensions adds capabilities for providing collections that are not free of charge and require purchasing data products that can thereupon be used in processing. Commercial data collections usually allow purchasing small subsets of the data (products), for example a single observation of an area.
+
+Therefore, the client must have an ability to search the available products that match their desired criteria and inspect their metadata to decide whether the products should be purchased.
+
+The client can then create an order for the desired products. Because of the financial cost of purchasing the data, a separate endpoint for confirming the execution of that specific order should be implemented.
+
+When the order is completed, the user should be able to access the data using the `collection-id` at the temporal and spatial location of the purchased products.
```
m-mohr, line 19: I think the parameter is generally called collection_id instead of collection-id in the API spec?!

```suggestion
When the order is completed, the user should be able to access the data using the `collection_id` at the temporal and spatial location of the purchased products.
```

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+}
+```
+
+### Ordering products
+
+Backends should implement the following endpoints:
+
+- `GET /orders`: Get a list of all created orders
+- `POST /orders`: Create an order
+- `GET /orders/{order-id}`: Get full metadata of a specific order
```
m-mohr, line 180: The parameter should be order_id instead of order-id, I think.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+When an order is created, the data isn't yet ordered from the commercial data provider. The client should explicitly confirm the order whereupon the order is executed, the costs are deducted from the client's account and the data is ingested in the associated collection. The `order:status` changes to `ordered`.
+
+If the user doesn't have sufficient funds, the endpoint should return an error and `order:status` should not change.
+
+### Payment
+
+Payment should be done in the currency used by the backend, listed at `GET /` under `billing`. When an order is created, the backend should return the full final cost of the order. 
+
+### Example usage with Python client
```
m-mohr, line 363: I haven't closely checked the Python example yet, I think @soxofaan or another Python client dev should also have a look.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+    - parameters
+  properties:
+    collection_id:
+      type: string
+      description: Unique identifier of the collection.
+    products:
+      type: array
+      description: Array of IDs of products to order.
+      items:
+        $ref: 'https://raw.githubusercontent.com/Open-EO/openeo-api/1.0.0/openapi.yaml#/components/schemas/order_id'
```
m-mohr, line 273: In the OpenAPI document, you should better use relative links.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+      4,
+      16
+    ]
+  },
+  "collection_id": "PLEAIDES",
+  "costs": 42
+}
+```
+
+#### `POST /orders/{order-id}`
```
m-mohr, line 353: What happens if I call this twice or so?

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+#### `GET /orders/{order-id}`
+
+Get full metadata of the order. The item should follow the [STAC Order Extension](https://github.com/stac-extensions/order), but extended with the spatial and temporal extent information and other metadata about the products.
+
+```json
+{
+  "order:id": "40264b5-c3ae-46f4-a907-0f612d763d97",
+  "order:status": "delivered",
+  "order:date": "2017-01-01T09:32:12Z",
+  "products": [
```
m-mohr, line 291: Instead of inlining them, we may want to consider linking to them. Also, I'm wondering whether we should generally also think about which of our additions we can propose to be added to the STAC extension?

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+- Version: **0.1.0**
+- Stability: **experimental**
+
+**Note:** This document only documents the additions to the specification.
+Extensions can not change or break existing behavior of the openEO API.
+
+## Overview of the workflow
+
+All the available datasets provided by a backend are listed on the `/collections` endpoint. The collections are normally freely accessible. This extensions adds capabilities for providing collections that are not free of charge and require purchasing data products that can thereupon be used in processing. Commercial data collections usually allow purchasing small subsets of the data (products), for example a single observation of an area.
```
m-mohr, line 13:
```suggestion
All the available datasets provided by a backend are listed on the `GET /collections` endpoint. The collections are normally freely accessible. This extensions adds capabilities for providing collections that are not free of charge and require purchasing data products that can thereupon be used in processing. Commercial data collections usually allow purchasing small subsets of the data (products), for example a single observation of an area.
```

m-mohr:
In general this looks pretty good. I've left some comments, some for discussion and some are usually minor fixes.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,398 @@
...
+  "assets": {...},
+  "order:status": "orderable",
+  "order_parameters": [
+      {
+        "name": "sample_type",
+        "description": "sample type of the output raster.",
+        "schema": {
+          "type": "string"
+        },
+        "enum": [
```
m-mohr, line 58: The enum should go into the schema, right?

m-mohr: Yeah, if `order_parameters` should basically be defined as `process_parameters` in `GET /service_types`, then you'd put the enum into the schema.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,398 @@
...
+          ],
+          [
+            12.36555287044679,
+            41.94403289260048
+          ]
+        ]
+      },
+      "properties": {
+        "constellation": "PHR",
+        "acquisitionDate": "2022-03-21T10:11:15.055Z",
```
m-mohr, line 305:
```suggestion
        "datetime": "2022-03-21T10:11:15.055Z",
```
That's how STAC defines and requires the acquisition date to be.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,398 @@
...
+        "azimuthAngle": 179.9852862071639,
+        "cloudCover": 0,
+        "geometryCentroid": {
```
m-mohr, line 308: Better use proj:centroid

extension/commercial_data/README.md
```diff
@@ -0,0 +1,398 @@
...
+            12.36555287044679,
+            41.94403289260048
+          ]
+        ]
+      },
+      "properties": {
+        "constellation": "PHR",
+        "acquisitionDate": "2022-03-21T10:11:15.055Z",
+        "azimuthAngle": 179.9852862071639,
+        "cloudCover": 0,
```
m-mohr, line 307: Better use eo:cloud_cover

extension/commercial_data/README.md
```diff
@@ -0,0 +1,398 @@
...
+        "geometryCentroid": {
+          "lat": 41.903935647240964,
+          "lon": 12.486569672582828
+        },
+        "processingLevel": "SENSOR",
```
m-mohr, line 312: Better use processing:level

extension/commercial_data/README.md
```diff
@@ -0,0 +1,398 @@
...
+  "parameters": {
+    "bounds": [
+      3,
+      15,
+      4,
+      16
+    ]
+  },
+  "collection_id": "PLEAIDES",
+  "costs": 42
```
m-mohr, line 329: This should go into properties, I think.

extension/commercial_data/README.md
Ardweaden: There is `payment`, but I that `Indicates a resource where payment is accepted.`.

It's strange that a relation type that just links to "additional information" does not exist. If we wanted to be more specific, I guess we could add `pricing-info | URL to human-readable pricing information.` or something like that?

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+Therefore, the client must have an ability to search the available products that match their desired criteria and inspect their metadata to decide whether the products should be purchased.
+
+The client can then create an order for the desired products. Because of the financial cost of purchasing the data, a separate endpoint for confirming the execution of that specific order should be implemented.
+
+When the order is completed, the user should be able to access the data using the `collection-id` at the temporal and spatial location of the purchased products.
+
+### Collection discovery
+
+A backend should add general information about a commercial data collection to the `/collections` and `/collections/{collection-id}` endpoints, the same as with freely available collections. Only the metadata about the entire dataset needs to be provided, not about the specific data products that a user has already purchased.
```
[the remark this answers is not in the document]
Ardweaden: I'm not sure this is needed, because the temporal and spatial availability of products is available in the information about orders.

However, this might be an argument for implementing `collections/{collection_id}/orders` as mentioned in https://github.com/Open-EO/openeo-api/pull/447#discussion_r924801576.

extension/commercial_data/README.md
Ardweaden: Added.

extension/commercial_data/README.md
Ardweaden: Thanks, fixed.

extension/commercial_data/README.md
Ardweaden: Thanks, fixed.

extension/commercial_data/README.md
Ardweaden: Thanks, fixed.

extension/commercial_data/README.md
Ardweaden: Thanks, fixed.

extension/commercial_data/README.md
Ardweaden: Done so.

extension/commercial_data/README.md
Ardweaden: Thanks, fixed.

extension/commercial_data/README.md
```diff
@@ -0,0 +1,416 @@
...
+### Searching available products
+
+Backends should implement top-level `GET /search` endpoint as specified in [STAC Item Search API specification](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/item-search). This should include the [Filter Extension](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-rc.1/fragments/filter), which enables filtering the available products by attributes specified in `/queryables` and `collections/{collection-id}/queryables`. 
+The endpoint should return a list of  `STAC Item`s that match the criteria.
+
+#### Example
+
+Example request payload to `GET /search` for `PLEIADES` products from "Living Library" with no snow coverage.
+
+```json
```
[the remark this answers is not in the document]
Ardweaden: Did the same for some others as well.

extension/commercial_data/README.md
m-mohr: Thinking more about it, we really should list the products at /collections/{collection_id}/items and just link to them from here so that we don't need to define the whole STAC structure again for this endpoint.

m-mohr:
@Ardweaden Thanks for the updates. I added another round of comments. ;-)

extension/commercial_data/README.md
```diff
@@ -0,0 +1,398 @@
...
+  "pricing": {
+    "searching": null,
+    "ordering": {
+      "description": "Minimum area per order is 0.25 km2. The price is calculated based on a 6-months sliding window.",
+    }
+  },
+  "links": [
+    {
+      "title": "Airbus Pleiades pricing",
+      "rel": "related",
```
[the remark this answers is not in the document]
m-mohr: See also https://github.com/Open-EO/openeo-api/pull/447#discussion_r924787502

extension/commercial_data/README.md
m-mohr: See also https://github.com/Open-EO/openeo-api/pull/447#discussion_r924801576

extension/commercial_data/README.md
Ardweaden: I changed all of this according to the suggestions, but I should clarify that this is an example of an Item as it is returned directly by Sentinel Hub, and that different products from different providers have various other fields, which are not present in this example.

extension/commercial_data/README.md
Ardweaden: Fixed.

extension/commercial_data/README.md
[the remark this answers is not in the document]
Ardweaden: This was added.

extension/commercial_data/README.md
Ardweaden: This would be quite a problem for our implementation, as Sentinel Hub doesn't support accessing information about a specific product directly, or listing all of them for that matter. It only supports searching them (which behind the scenes uses the searching functionality of the third parties who provide the data). 

In the "previous" proposal, `GET /orders` did not include any "heavy" information  about products in order to make it lightweight: https://github.com/Open-EO/openeo-api/pull/447/commits/27a806cc44ddb4e8bc1551c37b995464f6569046. This was intended in this proposal as well, perhaps I should clarify that. The products are listed inline only in `GET /orders/{order_id}`.

I would suggest we add `/items` and linking them as optional, and allow listing them inline by default.

extension/commercial_data/README.md
```diff
@@ -272,6 +272,7 @@ Get full metadata of the order. The item should follow the [STAC Order Extension
     {
       "type": "Feature",
       "stac_version": "1.0.0",
+      "stac_extensions": ["proj", "eo", "view", "processing"],
```
m-mohr, line 275: These are the v0.9 extension names, in v1.0.0 you need the schema URIs. For example `https://stac-extensions.github.io/eo/v1.0.0/schema.json` instead of `eo`. Please check the extension READMEs for the corresponding URIs.

extension/commercial_data/README.md
m-mohr: Okay, then the spec should probably highlight that providers are recommended to align with STAC as much as possible and not only dump existing information "as-is" as then it's not very useful.

extension/commercial_data/README.md
m-mohr: But isn't listing a list of products for a specific collection basically the same as searching all products for a given collection ID? The same applies for the Item ID, I'd think.

mkadunc: While I agree with @m-mohr that we should aim to reuse the existing STAC interfaces as much as possible, our implementation will inevitably have to take into account some limitations, specifically:

When listing `orderable` items:
- listing collection items for non-ordered products is only possible using Sentinel Hub's [3rd party data searching request](https://docs.sentinel-hub.com/api/latest/reference/#operation/dataImport_searchData);
- the generic part of the simple search API is _very_ limited (the only query parameters one can use in a generic way are `bounds`, `timeRange` and `maxCloudCoverage`);
- there are more options in the provider-specific "native search", but even these **don't currently allow fetching products by their ID** (except maybe with Planet data, where the native search is the most sophisticated);
- you always get all item data inline; there's paging to make 100+ products use-case manageable.

When listing `ordered` items:
- we can list all `product IDs` that were used as input when creating [a specific order](https://docs.sentinel-hub.com/api/latest/reference/#operation/dataImport_getOrder)
- after delivery, the delivery-related [endpoints](https://docs.sentinel-hub.com/api/latest/reference/#operation/dataImport_getOrderDelivery) again list only files and product IDs, but they don't contain STAC item metadata for the delivered products
- STAC item metadata is only available when the images are ingested into user collections (BYOC) within Sentinel Hub, and accessible via the specific collection's [catalog API](https://docs.sentinel-hub.com/api/latest/data/byoc/#catalog-api-capabilities)

My hope is that we can come up with a commercial data extension that is both reasonable as a standard interface as well as not-too-difficult to implement with the above constraints.

extension/commercial_data/README.md
Ardweaden: Added a clarification.

extension/commercial_data/README.md
Ardweaden: Thanks, fixed.

extension/commercial_data/README.md
Ardweaden: > the generic part of the simple search API is very limited (the only query parameters one can use in a generic way are bounds, timeRange and maxCloudCoverage);

We already have `/search` in this spec, so I think the searching functionality of `/items` isn't really needed, and I also don't see a big utility in listing all available products.

IMO the only really useful thing is to have information about a single item on a separate endpoint, but as you mentioned this is exactly what we don't support atm. That's why I proposed to have it optional.

What we could also to is to have e.g. `GET /orders/{order_id}/products` or `GET /orders/{order_id}/items` (with pagination) and list all the products  metadata there.

m-mohr: Given the limitations, I have two ideas:
1. Treat orders as new collections, i.e. each order creates a new collection at /collections and has then additional metadata for this specific order. You can then load specific orders via load_collection etc, but may end up with a lot of collections. You could then access the new data in two ways:
    1. via the original collection where the data for all orders can be accessed via load_collection.
    2. via the new collection where only data for the specific order can be accessed via load_collection
2. Do the same as above, but keep the new collections separate at /orders. Still, they should be full collections so that we can re-use as much code as possible from the collection implementations in the clients and back-ends, they just have some additional order-related metadata included.

Whether to choose 1 or 2 is more a "strategic" question on whether we want new orders to pop up in the collections endpoint or not.

Generally (i.e. for 1 and 2), no product metadata is inlined and you access them via .../items. Back-ends are free to choose whether they can make that only available for the new collections or also the old ones. You can give some information about products in that order, e.g. by listing product ids in collection summaries or so.

I think that should work having the limitations in mind, but maybe I missed something.

openapi.yaml
m-mohr: We could document both here so that users know that they exist or at least mention the global one. But the more important one is likely the collection-specific one, indeed.

extension/commercial_data/README.md
m-mohr: Resolving this in favor of https://github.com/Open-EO/openeo-api/pull/447#discussion_r924812905

extension/commercial_data/README.md
Ardweaden: > You could then access the new data in two ways:
> i. via the original collection where the data for all orders can be accessed via load_collection.
> ii. via the new collection where only data for the specific order can be accessed via load_collection

Doing something like this, i.e. saving purchased data to separate collections but allow loading them as one, is possible for us, as we have a "data fusion" feature, but if I'm not mistaken, requests to i.) would become linearly more expensive as new orders are added, so I don't think it's preferable (and it's definitely more complex). 

IMO it would be preferably to:

1. Have a single orderable "general" collection that contains data for all the orders and is used in `load_collection`, e.g. `collections/PLEIADES`
2. (Orderable) collections have `/items` where all ordered items are listed, e.g. `collections/PLEIADES/items`
3. I guess we will have to save the product Items on our backend and implement `collections/PLEIADES/items/{item-id}`, but only for the products used in orders, and not generally.
4. `orders/{order-id}` then use links to items

> I think that should work having the limitations in mind, but maybe I missed something. Happy to jump on another quick call to discuss details.

I would be happy to, when do you have time?

m-mohr: Today 13:00, 16:00, tomorrow 11:00, 13:00, > 15:00, Fri 13:00

Ardweaden: Today at 16:00 is fine for me, I think you should have received the invite.

Ardweaden: Sorry about the meeting time limit, it's maximum by Webex on my account.

> Retrieve order information: GET /orders/:id -> contains metadata about the order (source and target collection id [might be the same or different], order extension fields, optionally links to product metadata), no inline product metadata

+ array of ordered `product-id`s

> Optional: Retrieve product metadata: GET /collections/:id/items and GET /collections/:id/items/:item_id -> may return error if not ingested yet, otherwise list of items. The items should contain links to the respective orders that made them available.

Ok let's go with this, as it makes the most sense.
My problem is just the Sentinel Hub implementation, as I don't think it is possible for items in `GET /collections/:id/items` to equal the ones the user queried with `/search` and ordered, with existing Sentinel Hub services (`GET /collections/:id/items` would contain only basic information about spatial and temporal extent).

However, this solves the issue of linking orders to items very nicely.

mkadunc: > Can orders (in the future?) be created for multiple collections? e.g. ordering PLANET and MAXAR data at once?

I don't see this as a high priority requirement - cross-provider search might be useful sometimes, but cross-provider ordering seems too complicated to implement for not-so-much gain.

The interface would not need to change much, though - the `/order` endpoint is top-level, so we'd just need to include `collection_id` per each item and not have just one separate collection_id parameter. And we'd have to support multiple-output-collections-per-order. But I still suggest to leave this for some future upgrade.

m-mohr: The question about cross-provider orders was actually related to the question whether /orders or /collections/:id/orders would be preferrable. And I now used /orders above to have exactly this flexibility, otherwise we could also go with /collections/:id/orders, I think

mkadunc: > ... we could also go with /collections/:id/orders, I think

Even then, I think it's better to have `/orders`, so that all the user's orders are available in one place via  `GET /orders`. If one would want just a subset, we could at some point add filtering to this endpoint.

Ardweaden: I updated the README.md and I added the openAPI file for `/orders` and `/orders/{order_id}` endpoints. Should I add `/search`, `/items`, `queryables` etc. as well? Sadly it's not possible to reference them directly from Github because they contain relative references, so I'd have to copy most of it over.

m-mohr: I'll have a look next week, thanks. I don't think we necessarily $ref the other endpoints as long as they are mentioned (and linked to) in the README.

jdries:
Did a quick review. The part about searching for products seems to be quite independent of commercial data, and has some overlap with property filtering in load_collection. Could that also be a separate extension? (Or part of core?)

jdries:
Ah yes, so for this extension, the optional support for search becomes a harder requirement?
What about the alignment between /queryables and the collection property filtering that was based on summaries?

m-mohr:
For ordering you need to provide specific products/items and the only reasonable way to do this is probably search. So yes, this is at least recommended to implement (we don't really require any endpoints in the spec though, you just expose what you offer via `GET /` so that's why I not call it a hard requirement).

> What about the alignment between /queryables and the collection property filtering that was based on summaries?

I'm not sure. Are load_collection filtering options and available filters in search always the same? The current draft has `openeo:property_filters` and `summaries` for this purpose, but I'd also be fine to use /queryables for this if the question can be answered "yes".

jdries:
@m-mohr  I gave it some thought, and can't really come up with an example of a property that is only relevant in one case and not in the other. (load_collection vs search)

All common properties that we see being used often (orbit direction, relative orbit, tile id, cloud cover, instrument mode,...) are all relevant in both cases.

It would be very nice if a user can construct a query for product search, and transfer that to load_collection without requiring much adaptation, although not sure if that would still be possible as we already use openEO style callbacks for property filtering in load_collection right...

m-mohr:
The only one that comes to mind is likely the collection ID, which doesn't make sense as property filter.

CQL Text is probably something we could allow as alternative, but CQL JSON is as complex as what we have in openEO so there's not a lot of benefit, I'd assume. Especially as I don't know of a lot of existing "clients" that generate CQL in a user-friendly way (and in a way that you can simply copy it).

This is really getting off topic right now, let's continue in another issue if there's further demand.

extension/commercial_data/openapi.yaml
```diff
@@ -0,0 +1,328 @@
+openapi: 3.0.2
+info:
+  title: openEO Commercial data API extension
+  version: 0.1.0
+  description: 
+    $ref: README.md
```
m-mohr, line 6: This is unfortunately invalid although redoc handles it quite nicely. The linter and other tools don't like it because it is not allowed in the OpenAPI spec and they expect an OpenAPI doc and not a Markdown file.

I'd propose to change to
```yaml
openapi: 3.0.2
info:
  title: openEO Commercial data API extension
  version: 0.1.0
  description: >-
    The Commercial Data API extension provides an interface for discovering,
    ordering and using commercial data in the openEO API.
  contact:
    name: openEO Consortium
    url: 'https://openeo.org'
    email: openeo.psc@uni-muenster.de
  license:
    name: Apache 2.0
    url: 'http://www.apache.org/licenses/LICENSE-2.0.html'
externalDocs:
  url: README.md
tags:
  - name: Orders
    description: Management of commercial data orders
```

m-mohr:
Hey @Ardweaden,

I've worked on this today and made a couple of changes to the OpenAPI document mostly (and removed the endpoint descriptions from the README as they are now fully included in the OpenAPI file). Could you have a look at whether the changes all make sense to you, too? Most notable changes are likely the split of order:id into order:id and id, rename of products to item and moving "item_metadata" links into "links".

I also merged with the draft branch and moved the extension to the correct folder.

If you have any questions on changes etc, please let me know.

Ardweaden:
@m-mohr 
Thanks a lot.

Looks good on the first glance. I am implementing this spec at the moment, so I'd prefer to continue with merging when that is more or less finished, in case I come across some practical problems that would require changing it.

m-mohr:
Merging doesn't mean we can't change anymore, we can always make new PRs to refine it, too. :-)

m-mohr:
@Ardweaden The STAC order extension has been released. A couple of minor changes have been made before the release, e.g. the status codes are a bit different now. Can you update the PR to reflect the changes? You can now also add the corresponding entry to stac_extensions for the order extension. Thanks :-)

Ardweaden:
> @Ardweaden The STAC order extension has been released. A couple of minor changes have been made before the release, e.g. the status codes are a bit different now. Can you update the PR to reflect the changes? You can now also add the corresponding entry to stac_extensions for the order extension. Thanks :-)

Thanks, updated. What did you mean with "corresponding entry to stac_extensions"? Should the orders, which are not STAC Items, but use the fields from Order extension specify `stac_extensions`?

m-mohr:
No, just in actual STAC resources. Looks fine, I think. It was already included.

extensions/commercial-data/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+The client can then create an order for the desired products from a orderable collection. Because of the financial cost of purchasing the data, a separate endpoint for confirming the execution of that specific order should be implemented.
+
+When the order is completed, the data is ingested in a collection and its ID is available at `/orders/{order_id}` as `target_collection_id`. The user should be able to access the data using the `target_collection_id` at the temporal and spatial location of the purchased products.
+
+### Collection discovery
+
+A backend should add general information about a commercial data collection to the `/collections` and `/collections/{collection_id}` endpoints, the same as with freely available collections. Only the metadata about the entire dataset needs to be provided, not about the specific data products that a user has already purchased. 
+
+Commercial data collections are distinguished from freely available collections by including `"order:status": "orderable"` as specified in the [STAC Order specification](https://github.com/stac-extensions/order/tree/v1.0.0).
```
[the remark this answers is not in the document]
Ardweaden: Fixed.

dthiex (approved):

PR #457, opened 2022-07-19
Clarify the message field in logs

### LINKED ISSUE #455: support for multi-line log messages
When a batch job fails in the VITO back-end, the error logs typically contains a multi-line stack trace.
We currently encode the newlines (JSON style) as `\n` in the message string.
This `\n` is currently an ad-hoc solution in terms of the openEO API, because it does not specify how things  (like multi-line log messages) should be encoded.
Can we standardize this in some way, so that all clients can build on this?

E.g. the web editor component collapses all whitespace (newlines and indentation), which makes user support painful. 
And in the python client you also have to be careful to get a useful render of the log message.
Screenshot from 2022-06-08 17-28-12

m-mohr:
Fixes #455

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -6038,7 +6038,14 @@ components:
           $ref: '#/components/schemas/log_level'
         message:
           type: string
-          description: A message explaining the log entry.
+          description: >-
+            A concise message explaining the log entry.
+
+            Messages do *not* explicitly support [CommonMark 0.29](http://commonmark.org/)
+            syntax as other descriptive fields in the openEO API do,
+            but the messages may contain line breaks or indentation.
+
+            It is NOT RECOMMENDED to add stacktraces to the `message`.
           example: >-
             Can't load the UDF file from the URL `https://example.com/invalid/file.txt`.
             Server responded with error 404.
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -6037,7 +6037,14 @@ components:
           $ref: '#/components/schemas/log_level'
         message:
           type: string
-          description: A message explaining the log entry.
+          description: >-
+            A concise message explaining the log entry.
+
+            Messages do *not* explicitly support [CommonMark 0.29](http://commonmark.org/)
+            syntax as other descriptive fields in the openEO API do,
+            but the messages may contain line breaks or indentation.
```
soxofaan, line 6045: suggestion:

> ... MAY may contain line breaks or indentation, to be rendered as such.

soxofaan (approved):
Approved with minor suggestion

openapi.yaml
m-mohr:
```suggestion
            but the messages MAY contain line breaks or indentation.
```

I don't think we should add a requirement here that client should render them as such. We don't require formatting related capabilities anywhere else, I think.

PR #458, opened 2022-07-19
Several improvements for estimates, budget and costs #443

### LINKED ISSUE #443: Job estimates for UDFs etc
Some job estimates may not be possible, e.g. for a UDF. For UDFs a back-end may not be able to give an estimate as the code is not really known/understood by the back-end automatically. Thus, we should clarify in the API how a back-end should reject estimates in such cases (e.g. a specific error code via HTTP 400)

m-mohr:
Several improvements around estimates, costs and budgets:
- Don't require that the estimated costs are the upper limit. Services may specify the costs more freely depending on their terms of service.
- If a batch job can't be estimated reliably, a `EstimateComplexity` error should be returned. Fixes #443
- `GET /me`: Clarify the behavior of the field `budget`, which was incorrectly inherited from the resource budget field and thus the description did not make a lot of sense.
- Minimum value for resource `costs` and `budget` is 0.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2977,7 +2977,7 @@ paths:
 
         Back-end providers MAY specify an expiry time for the estimate.
         Starting to process data afterwards MAY be charged at a higher cost.
-        Costs MAY NOT include downloading costs. This can be indicated with the
+        Costs do often not include download costs. Whether download costs are included or not can be indicated explicitly with the
         `downloads_included` flag.
         
         The estimate SHOULD be the upper limit of the costs, but back-end are
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -2974,12 +2974,20 @@ paths:
       description: >-
         Calculates an estimate for a batch job. Back-ends can decide to
         either calculate the duration, the costs, the size or a combination of
-        them. This MUST be the upper limit of the incurring costs. Clients can
-        be charged less than specified, but never more. Back-end providers MAY
-        specify an expiry time for the estimate. Starting to process data
-        afterwards MAY be charged at a higher cost. Costs MAY NOT include
-        downloading costs. This can be indicated with the `downloads_included`
-        flag.
+        them.
+
+        Back-end providers MAY specify an expiry time for the estimate.
+        Starting to process data afterwards MAY be charged at a higher cost.
+        Costs MAY NOT include downloading costs. This can be indicated with the
+        `downloads_included` flag.
+        
+        The estimate SHOULD be the upper limit of the costs, but back-end are
+        free to use the field according to their terms of service.
+        
+        For some batch jobs it is not (easily) possible to estimate the costs
+        reliably, e.g. if a user-submitted UDF is part of the process.
+        In this case, the server SHOULD return a `EstimateComplexity` error
+        with HTTP status code 500.
       tags:
         - Data Processing
         - Batch Jobs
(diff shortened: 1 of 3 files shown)


openapi.yaml
```diff
@@ -2973,12 +2973,20 @@ paths:
...
-        be charged less than specified, but never more. Back-end providers MAY
-        specify an expiry time for the estimate. Starting to process data
-        afterwards MAY be charged at a higher cost. Costs MAY NOT include
-        downloading costs. This can be indicated with the `downloads_included`
-        flag.
+        them.
+
+        Back-end providers MAY specify an expiry time for the estimate.
+        Starting to process data afterwards MAY be charged at a higher cost.
+        Costs MAY NOT include downloading costs. This can be indicated with the
```
soxofaan, line 2980: What kind of downloading is meant here? Downloading raw/commercial data from a data source to the back-end or downloading the result from back-end to client?

openapi.yaml
```diff
@@ -2973,12 +2973,20 @@ paths:
...
+        `downloads_included` flag.
+        
+        The estimate SHOULD be the upper limit of the costs, but back-end are
+        free to use the field according to their terms of service.
+        
+        For some batch jobs it is not (easily) possible to estimate the costs
+        reliably, e.g. if a user-submitted UDF is part of the process.
```
soxofaan, line 2987: "user-submitted UDF" sounds like a pleonasm

openapi.yaml
```diff
@@ -3565,7 +3573,12 @@ paths:
                           to the user.
                         example: 1073741824
                   budget:
-                    $ref: '#/components/schemas/budget'
+                    type: number
+                    nullable: true
+                    description: |-
+                      The remaining budget a user has available.
+                      The value MUST be specified in the currency of the back-end.
+                      The value SHOULD be set to `null` if no explicit limit applies.
```
soxofaan, line 3581: > SHOULD be set to `null` if no explicit limit applies.

is that equivalent to omitting the field?

m-mohr: If it's not set, the meaning is pretty much undefined right now.

openapi.yaml
m-mohr: "Downloading raw/commercial data from a data source to the back-end" isn't covered by the API at all yet, thus there's #447. So the initial thinking was pretty much about "downloading the result from back-end to client".

```suggestion
        Costs MAY NOT include downloading costs from the back-end to the client. This can be indicated with the
```

openapi.yaml
m-mohr: I had to google "pleonasm" and I don't think I agree. I was thinking that in the future there could also be back-end provided UDFs which they might be able to give estimates for, although the question then is why it's not a pre-defined process, indeed. But I'd also be fine to remove the user-submitted here.

```suggestion
        reliably, e.g. if a UDF is part of the process.
```

openapi.yaml
soxofaan: > downloading the result from back-end to client"

I'm not completely sure I understand in what sense a back-end could/would charge for this on the job level. You can somewhat estimate the cost of a job on itself, but you don't know how many times or what subset of assets a user will download.

openapi.yaml
soxofaan: ok I understand.

Another example to consider here: doing ML prediction with a ML model that is "user-submitted" in the same sense (black box)

openapi.yaml
m-mohr: Yes, that is why it is saying it here. This is the cost for processing, likely not for any download that may happen for the results, which may incur additional costs.

m-mohr: True, MAY NOT is indeed not part of RFC2119. What I meant to say is something like:

```suggestion
        Costs do often not include download costs. Whether download costs are included can be indicated with the
```

PR #467, opened 2022-11-02
Clarify how the well-known document works #460

### LINKED ISSUE #460: well-known/openeo confusion
https://github.com/Open-EO/openeo-api/blob/16b1122c7090cc4fedcabace48686a4f73c1c876/openapi.yaml#L883-L886

While I was discussing something about `.well-known/openeo` with @dthiex  I found  this "URI MUST NOT be versioned" is a bit confusing: 
is it really "forbidden" to support a not-toplevel `.well-known/openeo`, e.g. something like `https://example.com/api/v1/.well-known/openeo`  or `https://example.com/foobar/.well-known/openeo`? 
Or should it just be a recommendation:   "URI SHOULD NOT be versioned" ?

m-mohr:
Fixes #460

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -921,25 +921,25 @@ paths:
 
         The Well-Known URI is the entry point for clients and users, so make
         sure it is permanent and easy to use and remember. Clients MUST NOT
-        require the well-known path (`/well-known/openeo`) in the URL that is
+        require the well-known path (`/.well-known/openeo`) in the URL that is
         specified by a user to connect to the back-end.
 
         For clients, the usual behavior SHOULD follow these steps:
         1. The user provides a URI, which may consist of a scheme (protocol), 
-           a authority (host, port) and a path.
+           an authority (host, port) and a path.
         2. The client parses the URI and appends `/.well-knwon/openeo` to the
            path. Make sure to correctly handle leading/trailing slashes.
         3. Send a request to the new URI.
            A. On success: Detect the most suitable API instance/version (see above)
               and read the [capabilites](#tag/Capabilities/operation/capabilities)
               from there.
-           B. On failure: Directly try to read the capabilities from the URI
+           B. On failure: Directly try to read the capabilities from the original URI
               given by the user.
 
-        **This URI MUST NOT be versioned as the other endpoints.** If your API
-        is available at `https://example.com/openeo/api/v1.0`, the Well-Known URI
-        SHOULD be located at `https://example.com/openeo/.well-known/openeo` and
-        the URI users connect to SHOULD be `https://example.com/openeo`.
+        **This URI MUST NOT be versioned as the other endpoints.**
+        If your API is available at `https://example.com/openeo/api/v1.0`, and
+        you instruct your API users to use `https://example.com/openeo` as connection URI, 
+        the Well-Known URI SHOULD be located at `https://example.com/openeo/.well-known/openeo`.
 
         Clients MAY get additional information (e.g. title or description) about
         a back-end from the most recent version that has the `production` flag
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -921,17 +921,25 @@ paths:
 
         The Well-Known URI is the entry point for clients and users, so make
         sure it is permanent and easy to use and remember. Clients MUST NOT
-        require the well-known path (`./well-known/openeo`) in the URL that is
-        specified by a user to connect to the back-end. A client MUST request
-        `https://example.com/.well-known/openeo` if a user tries to connect to
-        `https://example.com`. If the request to the well-known URI fails, the
-        client SHOULD try to request the capabilities at `/` from
-        `https://example.com`.
+        require the well-known path (`/well-known/openeo`) in the URL that is
```
soxofaan, line 924: > `/well-known/openeo`

is dropping the leading `.` intentional here?

openapi.yaml
```diff
@@ -921,17 +921,25 @@ paths:
...
+        specified by a user to connect to the back-end.
+
+        For clients, the usual behavior SHOULD follow these steps:
+        1. The user provides a URI, which may consist of a scheme (protocol), 
+           a authority (host, port) and a path.
```
soxofaan, line 929:
```suggestion
           an authority (host, port) and a path.
```

openapi.yaml
```diff
@@ -921,17 +921,25 @@ paths:
...
+        2. The client parses the URI and appends `/.well-knwon/openeo` to the
+           path. Make sure to correctly handle leading/trailing slashes.
+        3. Send a request to the new URI.
+           A. On success: Detect the most suitable API instance/version (see above)
+              and read the [capabilites](#tag/Capabilities/operation/capabilities)
+              from there.
+           B. On failure: Directly try to read the capabilities from the URI
```
soxofaan, line 936:
```suggestion
           B. On failure: Directly try to read the capabilities from the original URI
```

openapi.yaml
```diff
@@ -921,17 +921,25 @@ paths:
...
+              given by the user.
 
         **This URI MUST NOT be versioned as the other endpoints.** If your API
-        is available at `https://example.com/api/v1.0`, the Well-Known URI
-        SHOULD be located at `https://example.com/.well-known/openeo` and the
-        URI users connect to SHOULD be `https://example.com`.
+        is available at `https://example.com/openeo/api/v1.0`, the Well-Known URI
+        SHOULD be located at `https://example.com/openeo/.well-known/openeo` and
```
soxofaan, line 941: > If your API  is available at `https://example.com/openeo/api/v1.0`,  
> the Well-Known URI     SHOULD be located at `https://example.com/openeo/.well-known/openeo`

Stating it like that looks very arbitrary. Why not `https://example.com/.well-known/openeo` or `https://example.com/openeo/api/.well-known/openeo` ?

soxofaan: I think you should start from the usage instructions for users about which connection URI to use and append `/.well-known/openo` to that.

e.g.:

        **This URI MUST NOT be versioned as the other endpoints.** 
        If your API is available at `https://example.com/openeo/api/v1.0`, 
        and you instruct your API users to use `https://example.com/openeo` as connection URI, 
        the Well-Known URI SHOULD be located at `https://example.com/openeo/.well-known/openeo`.

openapi.yaml
soxofaan:
```suggestion
        require the well-known path (`/.well-known/openeo`) in the URL that is
```

m-mohr: No, thanks for catching this.

openapi.yaml
m-mohr: Yes, it was meant to be arbitrary because implementers can choose arbitrary paths. That was your criticism in #460, wasn't it? ;-)

I've updated the text according to your proposal.

soxofaan (approved):

Issue #276, opened 2020-04-23
Batch job options/configuration settings


soxofaan:
Currently in VITO backend we support (backend specific) batch job options for example to finetune spark driver and executor memory limits. These are provided in the `POST /jobs` request in a "job_options" field, e.g. something like this:

    POST /jobs
    {
        "process": { .... },
        "job_options": {"driver-memory": "10g", "executor-memory": "13g"},
         ....

This is also supported in the python client (e.g. see https://github.com/Open-EO/openeo-python-client/blob/1171465c69f85f3c225719a6664c7a3865076ff1/openeo/rest/connection.py#L463-L471)

The actual options and usage of this is of course highly backend specific, but it's probably a good idea to reserve this field "job_options" in official spec.

m-mohr:
The JSON object in the body of /jobs (and other similar endpoints) is intentionally open (i.e. additionalProperties is set to true) for such additions. Whether you add job_options or put for example driver-memory on the top-level is pretty back-end specific and thus I don't see any to do for the API spec.

soxofaan:
Also: I'm not pushing to include this in the 1.0.0 API. If there is consensus to include it in a future version, that's fine too.

m-mohr:
@soxofaan I just added a clarification to the client library guidelines that the `additional` parameter should be merged with the request body.

This behavior was actually foreseen in the API and was meant to be used as follows:

So for example:
`createJob(process, title, description, plan, budget, {"driver-memory": "10g", "executor-memory": "13g"})`
would lead to the following request:

```
POST /jobs
{
    "process": { .... },
    "driver-memory": "10g",
    "executor-memory": "13g",
     ....
```

It just lacked the actual wording that additional was meant to be merged with the request body.

If you want to stick with job_options on the back-end side, I'd propose to just use the Python client as follows:
`createJob(process, title, description, plan, budget, {"job_options": {"driver-memory": "10g", "executor-memory": "13g"}})`

That would make at least the clients compatible.

m-mohr:
The page build hasn't finished yet:

The addition is:
> * **additional** in `createJob` and `createService` (also below in `updateJob` and `updateService`): May hold additional key-value-pairs as object. The object should be merged with the body of the request.

soxofaan:
interesting, thanks

I think we'll stick with "namespacing" these job configuration things under a toplevel "job_options" field

m-mohr:
[...]

PR #471, opened 2022-11-15
Processing Parameters Extension #276 #545

m-mohr:
Adds the Processing Parameters Extension.

This extension adds a new endpoint (`GET /processing_options`)
to discover the additional processing options that a back-end offers.

Additionally, this extension allows to provide specific default values for user-defined processes (UDPs, see below).

Implements #276 and #545

### DIFF
--- a/extensions/processing-parameters/README.md
+++ b/extensions/processing-parameters/README.md
@@ -46,6 +46,20 @@ The values of the object are the default values for the parameters.
 Schematic restrictions are not defined for the object, but the schemas for the parameters as defined in `GET /processing_parameters` apply to the given values.
 These values provide the defaults unless a user overrides them in the actual data processing request (e.g. `POST /jobs`, see below).
 
+Example UDP including defaults for the processing parameters `memory` and `block-sizes` of a batch job:
+
+```json
+{
+  "id": "my_evi",
+  "parameters": [...],
+  "process_graph": {...},
+  "default_job_options": {
+    "memory": "2GB",
+    "block-sizes": [128, 32]
+  }
+}
+```
+
 ## Resolving parameters
 
 Due to the variety of places where processing parameters can be provided, the following
--- a/extensions/processing-parameters/README.md
+++ b/extensions/processing-parameters/README.md
@@ -36,8 +36,8 @@ additionalProperties:
   description: Any type
 ```
 
-The keys of the object are the respecive parameter names.
-The values of the object are the (default) values for the parameters.
+The keys of the object are the respective parameter names.
+The values of the object are the default values for the parameters.
 Schematic restrictions are not defined for the object, but the schemas for the parameters as defined in `GET /processing_options` apply to the given values.
 These values provide the defaults unless a user overrides them in the actual data processing request (e.g. `POST /jobs`, see below).
 
--- a/extensions/processing-parameters/README.md
+++ b/extensions/processing-parameters/README.md
@@ -9,3 +9,48 @@ The Processing Parameters Extension to the openEO API provides an interface to e
 
 **Note:** This document only documents the additions to the specification.
 Extensions can not change or break existing behavior of the openEO API.
+
+This extension adds a new endpoint (`GET /processing_options`, see [OpenAPI document](openapi.yaml))
+to discover the additional processing options that a back-end offers.
+
+Additionally, this extension allows to provide specific default values for user-defined processes (UDPs, see below).
+
+In both cases, the parameters and its values are provided separately for each processing mode.
+
+## Processing Parameters in UDPs
+
+UDPs can provide default values for specific processing parameters.
+
+The values for each parameter are provided  separately for each processing mode.
+The following properties are added to the top-level of a UDP for the respective processing modes:
+
+- `default_synchronous_parameters` for synchronous processing
+- `default_job_parameters` for batch jobs
+- `default_service_parameters` for secondary web services
+
+The schema for each of these properties is:
+
+```yaml
+type: object
+additionalProperties:
+  description: Any type
+```
+
+The keys of the object are the respecive parameter names.
+The values of the object are the (default) values for the parameters.
+Schematic restrictions are not defined for the object, but the schemas for the parameters as defined in `GET /processing_options` apply to the given values.
+These values provide the defaults unless a user overrides them in the actual data processing request (e.g. `POST /jobs`, see below).
+
+## Resolving parameters
+
+Due to the variety of places where processing parameters can be provided, the following
+list defines how the parameters must be resolved. The prioritization is as follows:
+
+1. If present, use the parameter specified in the processing request directly (e.g. in `POST /jobs` as a top-level property)
+2. If present, use the default parameter specified in the UDP
+3. Otherwise, use the default value for the parameter as specified in `GET /processing_options`
+
+"Present" means that the property is present in the JSON representation regardless of the value given, i.e.
+properties are present if an empty string, an empty array, an empty object, `false`, `0`, or `null` are provided.
+
+Unrecognized/unknown parameters that are provided through UDPs must be ignored by backends.
--- a/extensions/processing-parameters/README.md
+++ b/extensions/processing-parameters/README.md
@@ -1,0 +1,11 @@
+# Processing Parameters Extension
+
+The Processing Parameters Extension to the openEO API provides an interface to explore and handle additional processing options that a back-end can offer for the three processing modes (synchronous processing, batch jobs, secondary web services).
+
+- Version: **0.1.0**
+- Stability: **experimental**
+- [OpenAPI document](openapi.yaml)
+- Conformance class: `https://api.openeo.org/extensions/processing-parameters/0.1.0`
+
+**Note:** This document only documents the additions to the specification.
+Extensions can not change or break existing behavior of the openEO API.
--- a/extensions/processing-parameters/README.md
+++ b/extensions/processing-parameters/README.md
@@ -1,6 +1,6 @@
 # Processing Parameters Extension
 
-The Processing Parameters Extension to the openEO API provides an interface to explore and handle additional processing options that a back-end can offer for the three processing modes (synchronous processing, batch jobs, secondary web services).
+The Processing Parameters Extension to the openEO API provides an interface to explore and handle additional processing parameters that a back-end can offer for the three processing modes (synchronous processing, batch jobs, secondary web services).
 
 - Version: **0.1.0**
 - Stability: **experimental**
@@ -10,8 +10,8 @@ The Processing Parameters Extension to the openEO API provides an interface to e
 **Note:** This document only documents the additions to the specification.
 Extensions can not change or break existing behavior of the openEO API.
 
-This extension adds a new endpoint (`GET /processing_options`, see [OpenAPI document](openapi.yaml))
-to discover the additional processing options that a back-end offers.
+This extension adds a new endpoint (`GET /processing_parameters`, see [OpenAPI document](openapi.yaml))
+to discover the additional processing parameters that a back-end offers.
 
 Additionally, this extension allows to provide specific default values for user-defined processes (UDPs, see below).
 
@@ -21,12 +21,12 @@ In both cases, the parameters and its values are provided separately for each pr
 
 UDPs can provide default values for specific processing parameters.
 
-The values for each parameter are provided  separately for each processing mode.
+The values for each parameter (so called 'options') are provided separately for each processing mode.
 The following properties are added to the top-level of a UDP for the respective processing modes:
 
-- `default_synchronous_parameters` for synchronous processing
-- `default_job_parameters` for batch jobs
-- `default_service_parameters` for secondary web services
+- `default_synchronous_options` for synchronous processing
+- `default_job_options` for batch jobs
+- `default_service_options` for secondary web services
 
 The schema for each of these properties is:
 
@@ -38,7 +38,7 @@ additionalProperties:
 
 The keys of the object are the respective parameter names.
 The values of the object are the default values for the parameters.
-Schematic restrictions are not defined for the object, but the schemas for the parameters as defined in `GET /processing_options` apply to the given values.
+Schematic restrictions are not defined for the object, but the schemas for the parameters as defined in `GET /processing_parameters` apply to the given values.
 These values provide the defaults unless a user overrides them in the actual data processing request (e.g. `POST /jobs`, see below).
 
 ## Resolving parameters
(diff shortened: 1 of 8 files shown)


m-mohr:
I'm hesistant to add it there to keep it simple/small and not delay the request too much, because it's the first one and should not block additional requests because it needs to wait for some options to be retrieved. We also have /service_types externally for this reason.

soxofaan:
I get that it feels normal from the web editor point of view, because you probably list a user's jobs before they are able to create a new one.

But in case of less interactive or lazy client environments like python client usage:  it's a bit weird that the client first has to request `GET /jobs`   in order to  create a new job.

m-mohr:
No, I think you understood correctly, but I also assumed that with other endpoints and it did turnout that this was a false assumption.

m-mohr:
I've updated the PR to have a dedicated (optional) endpoint.

openapi.yaml
```diff
@@ -2386,6 +2386,41 @@ paths:
...
+  /processing_options:
+    get:
+      summary: Additional processing options
+      operationId: list-processing-options
+      description: |-
+        Lists additional custom processing options that a back-end offers for the
+        different processing modes (synchronous processing, batch jobs, secondary web services).
+
+        The options specified here can be added to the corresponding `POST` requests
+        at the top-level of the object sent in the payload.
```
soxofaan, line 2398: Maybe also make it clear that  all options, being "additional", should be optional (or have default values), 
in the sense  that a batch job will still be accepted when none of these options are specified at job create time.
Or differently put: a backend should not define a _required_ additional parameter.

m-mohr: A back-end could do that actually. Normally everything should be optional, but if one chooses to require something here, it is free to do so. But I'll add some text to give some more context...

dthiex (approved):

openapi.yaml
soxofaan: > but if one chooses to require something here, it is free to do so

Isn't that a breaking change in some way then?

m-mohr: For their implementation, it might be breaking, yes. Every implementation is free to break whatever they like in their implementation ;-) But that's independent of the API spec and for the API spec it's not a breaking change, no.

m-mohr: Added:

> All options SHOULD explicitly be made optional with reasonable defaults
> as otherwise the interoperability between the implementations decreases.

dthiex (approved):

m-mohr:
This PR only specifies what happens for POST requests (reflected in descriptions and field names `create_...`), but doesn't say anything about update. -> todo

openapi.yaml
```diff
@@ -2397,6 +2397,43 @@ paths:
...
+    get:
+      summary: Additional processing options
+      operationId: list-processing-options
+      description: |-
+        Lists additional custom processing options that a back-end offers for the
+        different processing modes (synchronous processing, batch jobs, secondary web services).
+
+        The options specified here can be added to the corresponding `POST` requests
+        at the top-level of the object that is sent as the payload.
+        All options SHOULD explicitly be made optional with reasonable defaults
```
soxofaan, line 2410: Couple of slightly related notes/questions:

- are the defaults global or can they depend on user (e.g. subscription level or some other property)?
- is it allowed to use a default like `null` in the public parameter specification which allows the backend to use a an actual fallback value that is not public or dynamic in some sense (e.g. user dependent, dependent on current cluster load, automatically derived from  spatial/temporal extent, ...)?
- is it necessary to cover these concerns in the description?

m-mohr:
Moved this from 1.2.0 to 1.3.0. Implementors are free to implement this proposal and give feedback, but it didn't feel fully fleshed out yet for a release in 1.2.0.

m-mohr:
As per https://github.com/Open-EO/openeo-api/issues/545#issuecomment-2517685847, this PR should be reworked to be an extension.

openapi.yaml
m-mohr: I think that's all up to the implementation to decide.

m-mohr:
This PR has been updated to include the functionality requested in #545 and has been migrated from an addition to the core API to an experimental extension to the API. @jdries @soxofaan

jdries (approved):
LGTM, we'll want to follow up with an implementation in our backend.

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+The schema for each of these properties is:
+
+```yaml
+type: object
+additionalProperties:
+  description: Any type
+```
+
+The keys of the object are the respecive parameter names.
```
soxofaan:
```suggestion
The keys of the object are the respective parameter names.
```

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+The values of the object are the (default) values for the parameters.
```
soxofaan:
```suggestion
The values of the object are the default values for the parameters.
```
I don't think there is a reason to put "default" between parentheses

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+## Processing Parameters in UDPs
+
+UDPs can provide default values for specific processing parameters.
+
+The values for each parameter are provided  separately for each processing mode.
+The following properties are added to the top-level of a UDP for the respective processing modes:
+
+- `default_synchronous_parameters` for synchronous processing
+- `default_job_parameters` for batch jobs
+- `default_service_parameters` for secondary web services
```
soxofaan: (feel free to ignore if this is too much bike-shedding)

I'm not sure if it is intentional and there is a reasoning behind it, but the current description mixes "parameter" and "option": "processing parameter extension", `GET /processing_options`, `default_job_parameters`, ...

To me, "parameter" sounds more like a non-concrete placeholder, while "option" sounds a bit more concrete (e.g. a value provided by the user to make a parameter concrete). So in that sense, I'd think that the naming should be switched to:
- `GET /processing_parameters`
- `default_synchronous_options`,  `default_job_options`, ....


I would also be fine with just collapsing everything to a single name, e.g. "option"

m-mohr: Indeed, it should just be a single name for it. Not sure yet which is better, but something to consider. Will have another look tomorrow.

soxofaan: Can we proceed on picking a single name for this concept? We're starting on implementing support for this, so it would be best if we can settle this detail.

I tend to prefer "option".
"Parameter" is already a defined concept in openeo process graph handling (including resolution and scoping rules), so I would avoid overloading it here with a somewhat different/incompatible meaning.

m-mohr: I was thinking into the same direction, but we have also named it paramerter pretty much everywhere else in the API, in services types, in file formats, ... (but we use options in the processes). In the programming clients they'd probably be exposed as parameters? Or as a single parameter called options?

Not sure, any other preferences?

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+- [OpenAPI document](openapi.yaml)
+- Conformance class: `https://api.openeo.org/extensions/processing-parameters/0.1.0`
+
+**Note:** This document only documents the additions to the specification.
+Extensions can not change or break existing behavior of the openEO API.
+
+This extension adds a new endpoint (`GET /processing_options`, see [OpenAPI document](openapi.yaml))
+to discover the additional processing options that a back-end offers.
+
+Additionally, this extension allows to provide specific default values for user-defined processes (UDPs, see below).
```
soxofaan: I don't think this extension has to be limited to user-defined processes. It also should apply to remote process definitions (https://github.com/Open-EO/openeo-api/tree/draft/extensions/remote-process-definition).

Ultimately, it could also apply to predefined processes. However, as mentioned at https://github.com/Open-EO/openeo-api/issues/545#issuecomment-2362487544), this would  be kind of an anti-pattern against the vision of openEO, so I don't think this should be (explicitly) in scope at this point.


```suggestion
Additionally, this extension allows to provide specific default values for processing values, embedded in process definitions, such as user-defined processes (UDPs, see below) and [remote process definitions](https://github.com/Open-EO/openeo-api/tree/processing-parameters/extensions/processing-parameters).
```

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+In both cases, the parameters and its values are provided separately for each processing mode.
+
+## Processing Parameters in UDPs
```
soxofaan:
```suggestion
## Embedding default processing options in process definitions
```

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+UDPs can provide default values for specific processing parameters.
```
soxofaan, line 27:
```suggestion
Process definitions (such as user-defined processes or remote process definition resources) can include default values for specific processing parameters.
```

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+The values for each parameter are provided  separately for each processing mode.
+The following properties are added to the top-level of a UDP for the respective processing modes:
```
soxofaan:
```suggestion
The following properties are added to the top-level of a process definition (e.g. as sibling nodes to the "process_graph" property) for the respective processing modes:
```

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+type: object
+additionalProperties:
+  description: Any type
+```
+
+The keys of the object are the respective parameter names.
+The values of the object are the default values for the parameters.
+Schematic restrictions are not defined for the object, but the schemas for the parameters as defined in `GET /processing_options` apply to the given values.
+These values provide the defaults unless a user overrides them in the actual data processing request (e.g. `POST /jobs`, see below).
+
```
soxofaan, line 48: Maybe add an example here to make things a bit more concrete:

```
Example process definition including processing option defaults:

{
    "id": "my_evi",
    "parameters": [...],
    "process_graph": {...},
    "default_job_parameters": {
        "memory": "2GB",
        "block-sizes": [128, 32]
    }
}
```

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,56 @@
...
+## Resolving parameters
+
+Due to the variety of places where processing parameters can be provided, the following
+list defines how the parameters must be resolved. The prioritization is as follows:
+
+1. If present, use the parameter specified in the processing request directly (e.g. in `POST /jobs` as a top-level property)
+2. If present, use the default parameter specified in the UDP
```
soxofaan, line 69:
```suggestion
2. If present, use the default parameter specified in the process definition
```

soxofaan (commented):
A couple of additional suggestions.

(some of them related to being a bit more general than UDPs)

extensions/processing-parameters/README.md
soxofaan: > file formats, ... (but we use options in the processes). In the programming clients they'd probably be exposed as parameters? Or as a single parameter called options?

Interesting I didn't notice that yet. In the openEO API, under `GET /file_formats` we speak about "parameters" (including the "parameter" property in a file format object), but in `save_result` we use argument name `options` (while its description still refers to "parameters" however). Was that an intended choice or did it just grow organically to that?

(FYI: in the Python client, we mainly expose the `options` argument of `save_result`, there is not really a prominent reference to "parameters")

soxofaan: > named it paramerter pretty much everywhere else in the API, in services types

under service types there is indeed "process_parameters", but that is directly tied to the classic parameters we have in process graph handling, so that's fine I think

soxofaan: The existing situation of file formats is an interesting precedent actually. While sticking to a single name might be easier for end users to connect the dots, it also makes sense to be bit more nuanced about the difference between defining a parameter and providing an actual value for it, e.g.
- the backend provides a file format with a _parameter_  (a non-concrete placeholder) for example "jpeg_quality"
- the user picks  _option_  `80%` (a concrete value) for jpeg quality


If we want to align with that existing pattern, then it would be here I think:
- name "Processing parameter extension"  (already the case in current PR)
- new endpoint `GET /processing_parameters`   (this is different)
- default values in UDPs and related: `default_synchronous_options`, `default_job_options`, `default_service_options`  (also different)

m-mohr: That makes sense to me.

extensions/processing-parameters/README.md
m-mohr: Remote process definitions are UDPs, too. I don't think I understand why the differentiation is needed.
Process definition is not a term in the API spec, but UDP is. So I'm inclined to leave it as is.

extensions/processing-parameters/README.md
m-mohr: See https://github.com/Open-EO/openeo-api/pull/471/commits/e9ed2bbcd0b32a5007523cef1433d272ef1dea9a

extensions/processing-parameters/README.md
m-mohr: Nevertheles, I clarified that there are various "sources" for such UDPs, including remote UDPs:
https://github.com/Open-EO/openeo-api/pull/471/commits/f7b765bf71272420e62019340102c08b111df5d9

extensions/processing-parameters/README.md
m-mohr:
```suggestion
## Embedding default processing options in UDPs
```

extensions/processing-parameters/README.md
m-mohr:
```suggestion
The following properties are added to the top-level of a UDP (e.g. as sibling nodes to the "process_graph" property) for the respective processing modes:
```

extensions/processing-parameters/README.md
m-mohr: See https://github.com/Open-EO/openeo-api/pull/471/commits/a407db56dcb22633f218622d547f7c97dc61f057

extensions/processing-parameters/README.md
```diff
@@ -0,0 +1,75 @@
...
+**Note:** This document only documents the additions to the specification.
+Extensions can not change or break existing behavior of the openEO API.
+
+This extension adds a new endpoint (`GET /processing_parameters`, see [OpenAPI document](openapi.yaml))
+to discover the additional processing parameters that a back-end offers.
+
+Additionally, this extension allows to provide specific default values for user-defined processes (UDPs),
+which includes:
+
+- UDPs submitted directly for synchronouzs processing, as batch jobs, or as secondary webservices
```
soxofaan:
```suggestion
- UDPs submitted directly for synchronous processing, as batch jobs, or as secondary web services
```

extensions/processing-parameters/README.md
soxofaan: ok fair enough. For me, the term "UDP" in the openEO API spec was always closely tied to the `/process_graphs` endpoints. But indeed, after checking in more detail, the openEO API  is a bit more relaxed about the term.

soxofaan:
apart from that typo fix https://github.com/Open-EO/openeo-api/pull/471#discussion_r1945125929, I'm fine for merging this already

PR #491, opened 2023-04-24
Federation extension: Add experimental and deprecated to back-end list

m-mohr:


jdries (approved):
I think this is in line with other openEO metadata, so looks good.
For backend maturity, we probably want to also add a more fine-grained system, like TRL levels: https://en.wikipedia.org/wiki/Technology_readiness_level but this can be part of another discussion/PR, as it also requires guidance in how it is interpreted in a specific federation.

PR #494, opened 2023-05-19
Crosswalk between openEO API and OCG API - Processes

m-mohr:
This document gives a brief overview over the similarities and differences between
- openEO API, v1.2.0
- OCG API - Processes - Part 1: Core, v1.0.0


jdries:
Maybe important to mention that openEO also defines the actual processes? I know this is not part of the api strictly speaking, but it is the key difference. The other one probably being how processes graphs can be chained and nested.

crosswalks/ogcapi-processes.md
```diff
@@ -0,0 +1,309 @@
...
+As the landing pages both are based on top of OGC API - Common, they are very similar.
+
+Some differences include:
+- openEO requires the following additional fields:
+  - Defined in OAP1, but not required: `title`, `description`
+  - Not defined in OAP1: `api_version`, `backend_version`, `stac_version`, `id`, `endpoints`, (`type`)
+- The relation type to link to the conformance classes is `conformance` in openEO (originating from
+  OGC API - Features / STAC API) and `http://www.opengis.net/def/rel/ogc/1.0/conformance` in OAP1.
+  Two links with the same target but different relation types will be required.
+- The existance of API endpoints in openEO is primarily checked through `endpoints`, which is
```
m-mohr, line 91: See https://github.com/Open-EO/openeo-api/issues/506

jdries:
I suggest to raise the point about 'data access' to the level of key differentiators.
Not being able to discover available data collections in a standardized manner is quite a big limitation. It means that an OGC Processes endpoint could expose 'load_collection', but there's no proper way to list available collections.

PR #502, opened 2023-07-03
 Crosswalk between openEO API and OCG API - Coverages

m-mohr:


crosswalks/ogcapi-coverages.md
```diff
@@ -0,0 +1,428 @@
...
+
+- OAC1: `GET /conformance` (required)
+
+Both endpoints are 100% equivalent (**openEO** additionally allows to list the conformance classes in the landing page,
+following the STAC API).
+
+
+### API description
+
+- openEO: `n/a` (?)
```
m-mohr: The way of linking to OpenAPI documents in openEO and OAC1 is the same, it's just that OAC1 pre-defines a path for it and requires the links, which are optional in openEO.

crosswalks/ogcapi-coverages.md
```diff
@@ -0,0 +1,428 @@
...
+
+Optionally, **openEO** allows for searching for individual STAC items "hidden"
+inside the collections, as per [STAC API repository](https://github.com/radiantearth/stac-spec/tree/v0.9.0/api-spec).
+
+
+### Data description
+
+- openEO:
+  - [`GET /collections/{collectionId}`](https://api.openeo.org/1.2.0/#tag/EO-Data-Discovery/operation/describe-collection) (required)
+  - [`GET /collections/{collectionId}/queryables`](https://api.openeo.org/1.2.0/#tag/EO-Data-Discovery/operation/list-collection-queryables) (required?)
```
m-mohr: it's optional.

crosswalks/ogcapi-coverages.md
```diff
@@ -0,0 +1,428 @@
...
+a wide range of benefits, **OAC1** provides a somewhat simpler shortcut
+for via query parameters of the URL.
+
+**OAC1** also allows for scaling on *any* coverage's dimension, while
+**openEO** only along the spatial plane/dimensions.
+
+
+### Tiles
+
+- openEO: `n/a`
```
m-mohr: openEO allows it through secondary web services, similar to OAC1.

crosswalks/ogcapi-coverages.md
```diff
@@ -0,0 +1,428 @@
...
+[OGC API - Tiles](http://docs.ogc.org/DRAFTS/20-057.html) standard, allows the retrieval of
+coverage data in form of tiles, hence trimmed and resampled to match the 
+boundaries and resolution of a given tile. More than one tilesets can be attached to a coverage.
+
+**openEO** does not provide tiled access directly, but can be easily cascaded to
+a dedicated secondary tiles service (like [XYZ tiles](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames))
+thanks to the [`/service_types`](https://api.openeo.org/1.2.0/#tag/Capabilities/operation/list-service-types) resource.
+
+
+### Media encoding
```
m-mohr, line 397: I'm not sure whether comparing these two is correct. Let's discuss tomorrow.

crosswalks/ogcapi-coverages.md
pierocampa: Yes, I mention that, but I'll make it clearer by replacing the n/a also.

crosswalks/ogcapi-coverages.md
```diff
@@ -0,0 +1,428 @@
...
+**`item`**
+A GeoJSON Feature augmented with foreign members relevant to a STAC object.
+
+**`cube:dimensions`**
+A set of dimensions that can be of different types: spatial, temporal, "other", etc.
+
+
+## Crosswalk
+
+*Executive summary (tl;dr)*
```
m-mohr, line 121: For me an important details is missing here. Can OGC API - Coverages be combined with the openEO API? It seems they have a common baseline and it could just be additional endpoints (mostly `/collections/{collectionId}/coverage/...`). 

Due to the openEO WCPS driver, we could translate openEO to WCPS, right? It seems vice-versa is not easily possible (waiting for an answer from Peter Baumann).

crosswalks/ogcapi-coverages.md
pierocampa: So in openEO there is no predefined link to the OpenAPI docs? How can a client discover the link?
(feel free to edit the document yourself anyway, of course)

m-mohr:
Thanks for the comparisong @pierocampa.

I've added a couple of comments above. I think an important question to answer here is how could openEO be combined with Coverages and/or WCPS. What would someone need to do to expose Coverages endpoints in openEO? Could WCPS and openEO processes somehow be translated?

crosswalks/ogcapi-coverages.md
m-mohr: In both specs, you use the relation types `service-desc` and `service-doc`. I think in both cases they can point anywhere, but OGC APIs propose to use `/api` as endpoint.

m-mohr: (It's better to first discuss, come to a common understanding and update afterwards :-) )

m-mohr:
An additional point that, I think, is important to add: The actual definitions of datacubes, dimensions etc. How do they compare? Can they be mapped 1:1? Are there any conflicts?

crosswalks/ogcapi-coverages.md
pierocampa: There are probably a couple of assumptions to be made regarding the cubes' dimensions, which makes openEO/OGC not 100% compatible, but probably cover most of the "real" coverages that are there anyway: I mean that there is not necessarily a 1:1 agreement and directional alignment between the dimensions of a cube and the CRS' axes (imagine I don't know, an oblique plane/tile in a 3D fishtank), which is a case that the OGC Coverage model can describe, while I am not sure about this with openEO. ?

Also, since the `datacube` STAC extension merges a coverage's *domain* and *range* spaces into the same "dimensions" container, there is the little "pain" to define the way to identify which dimensions of a datacube are what (spatial, temporal, other, band, etc), based on fragile assumptions on axis/dimensions labels?

Translating a (subset of) (the most basic, like filtering, scaling, etc) openEO processes to WCPS is feasible (are you talking about the [openEO Spring driver](https://github.com/Open-EO/openeo-spring-driver)?).

The other way around is probably also feasible, with the additional step of requiring to identify the category of the dimensions, as openEO processes are much more strongly typed in this sense, WCPS treats all dimensions equally and operations on them do not "discriminate".

pierocampa:
> An additional point that, I think, is important to add: The actual definitions of datacubes, dimensions etc. How do they compare? Can they be mapped 1:1? Are there any conflicts?

Right that one. As I was mentioning in the other comment, this is probably the key and most painful point to be solved to reach compatibility (and happiness)!

crosswalks/ogcapi-coverages.md
```diff
@@ -0,0 +1,427 @@
...
+
+Most notably, an OGC [*coverage*](http://myogc.org/go/coveragesDWG) &mdash;
+even though formally defined as a *feature*
+(and not a *collection* thereof) &mdash; is frequently an abstraction for a 
+*collection* of (generally, but not limited to, gridded) datasets/items
+(like time-series of satellite images for instance),
+so we can generally compare it with STAC [collections](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md),
+especially [datacubes](https://openeo.org/documentation/1.0/datacubes.html).
+
+### OGC
```
m-mohr, line 79: For better understanding, it might help to give examples here so that we can more easily map between openEO and Coverages. 

I'm trying to map right now, but I'm unsure:
- collection = collection?
- coverage/feature = data cube?
- range = dimension?
- domain = dimension labels?

crosswalks/ogcapi-coverages.md
m-mohr: I guess openEO could describe the 3D fishtank if there's a CRS for it?

I agree that the STAC data cube extension is a bit messy with regards to the type, but it's specified. type = spatial (with axis), type = temporal, type = ... (e.g. bands). At least internally it should be clear. "other" only really shows up if you add a "user-defined" dimension in the process graph.

I'm not specifically talking about any implementation, more what theoretically can be translated or whether there are any roadblocks.

openEO processes only really "discriminate" spatial axes as x and y need usually be used together, but I'm not sure how WCPS handles the special case of spatial?

crosswalks/ogcapi-coverages.md
pierocampa: The mapping is more like this:

- collection = collection
- coverage/feature = data cube / item / collection 
- range set = the actual data (pixels, voxels, n-xels?!)
- range type = the description of the data space (bands) = dimensions::bands
- domain set = dimensions::spatial, dimensions::temporal, dimensions::other

crosswalks/ogcapi-coverages.md
pierocampa: > I guess openEO could describe the 3D fishtank if there's a CRS for it?

Yes, but how can the orientation of the cube dimensions be specified? Even in the simple/common case when the cube is "aligned" with the CRS axes: we should not rely on the labels of the dimensions to infer its "direction". OGC uses the "offset vectors" for this.

> I agree that the STAC data cube extension is a bit messy with regards to the type, but it's specified. type = spatial (with axis), type = temporal, type = ... (e.g. bands). At least internally it should be clear. "other" only really shows up if you add a "user-defined" dimension in the process graph.

I don't think it is messy, it is clear! Only I personally prefer the distinction that the OGC Coverages do between domain and range, it makes it even clearer.

> openEO processes only really "discriminate" spatial axes as x and y need usually be used together, but I'm not sure how WCPS handles the special case of spatial?

Afaik WCPS does not handle any special spatial (nor temporal) case: all operators are completely agnostic of the type of dimensions they operate onto.

crosswalks/ogcapi-coverages.md
pierocampa: (anyway I'm a bit confused here by your mapping: STAC datacube can also be a STAC collection, right?)

crosswalks/ogcapi-coverages.md
m-mohr: Dimension labels in openEO have an order, but otherwise there's nothing like that, indeed.

> Afaik WCPS does not handle any special spatial (nor temporal) case

Well, it does, right? bands (range) seems separate from spatial/temporal (domain), while in openEO it's just all a dimension. What's so special about bands?

crosswalks/ogcapi-coverages.md
m-mohr: Well, I'd keep STAC aside for now, it's just the metadata, but doesn't really care in this context (I think) as we should just map the coverage and data cube concepts.

Your list makes sense to me, I'm just confused about the range type. What happend with the range type if the data is non-spectral? And why is other a domain set then?

crosswalks/ogcapi-coverages.md
pierocampa: Yep, I was talking indeed about a Coverage dimensions: no special case there, but yes, bands indeed are treated separately. To me it makes a whole lot of sense to have a domain, a "space", (or space-time) **where** the measurements lie, and then you have the data-space, which is **what** are the measurements about.

crosswalks/ogcapi-coverages.md
pierocampa: Range type is just metadata about the data: it says what is it (the label given, the unit of measure, a link to a formal description, etc (a SWE record). It is separated from the domain set because the coverage model separates them! The *where* and the *what*, as I said in the previous comment.

And this is because the coverage is defined as a *function* that links every point in the domain space to a point in the range space, like saying: `(y_1*, y_2*, y_3*) = f(x_1*, x_2*)`. You are essentially asking why having the vector of dimensions `x_i` and the vector of dependent variables `y_i` not merged together. That's the way I see it!

crosswalks/ogcapi-coverages.md
m-mohr: Okay, it's good to understand this, thanks.

crosswalks/ogcapi-coverages.md
m-mohr: Okay, I just think that dimensions of type other in openEO are often more a range rather than a domain. It could for example be a data cube x,y,t,indices where indices is of type other and contains ndvi, evi, ndsi etc. Then I think it should be a range in WCPS, right?

pierocampa: The indices should go in the coverage range, exactly. Though this is not always the case! I modeled 5D marrays for climate indices projections with "model" and "scenario" as additional "other" dimensions (then x,y,t) (see an example [here](http://saocompute.eurac.edu/rasdaman/ows?&SERVICE=WCS&VERSION=2.0.1&REQUEST=DescribeCoverage&COVERAGEID=amt_19712100_tnst_1x1km_year_cordexadj_qdm)): I would **not** want these "other" dimensions to go into a coverage range, but rather stay as dimensions (domain). In the range I might want to keep the values only, or maybe I could put value and uncertainty, etc.

PR #513, opened 2023-09-29
Issue #507: clarify that "canonical" job result link should include query parameters

### LINKED ISSUE #507: clarification: canonical (signed) batch job result URL + partial query param?
I'm  working on crossbackend job execution in aggeragtor (https://github.com/Open-EO/openeo-aggregator/issues/115).

One of the pieces of the current solution we aim for is leveraging "partial" result listings through the `partial` parameter e.g. `/jobs/{job_id}/results?partial=true`.
However we also require signed URLs through a "canonical" link in `/jobs/{job_id}/results`, e.g. possibly `https://whatever.com/azerty123` (where the `azerty123` is an  opaque blob that securely can be resolved to the original batch job result listing)

It is not completely clear how the `partial=true` should be included in the signed URL: should it be encoded automatically in the opaque blob, or should the user add it explicitly, e.g. `https://whatever.com/azerty123?partial=true`?

I guess it's the former, but we might have to clarify that a bit .

E.g. current description state:

> It is strongly recommended to add a link with relation type canonical, which points to this STAC document using a signed URL. This way the STAC metadata can be used by non-openEO clients without additional authentication steps.

e.g. 
> ... which points to this STAC document (including all query parameters like `partial`) using a signed URL...

soxofaan:
Issue #507: clarify that "canonical" job result link should include query parameters

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `GET /file_formats`: Base paramater on top of normal JSON Schema, not Process JSON Schema
 - `PATCH /services/{service_id}` and `PATCH /jobs/{job_id}`: Explicitly allow updating back-end specific properties (as in `POST`)
 - `GET /services/{service_id}` and `GET /jobs/{job_id}`: Explicitly allow listing back-end specific properties (as provided in `POST`)
+- `GET /jobs/{job_id}/results`: Clarify that "canonical" job result link should include query parameters
 
 ## [1.2.0] - 2021-05-25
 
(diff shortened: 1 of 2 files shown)


m-mohr:
Thinking a bit more about this, I think this was intentionally left out as implementations could either choose to include this parameter in the canonical link href or encode it implicitly in the signed URL. Depending on the implementation, it should or should not be part of the canonical href. Clients (which includes the aggregator) would basically just follow the link that was provided. If the href doesn't work as provided by the implementation, that's an implementation bug. 

Makes sense, right? If yes, could you update the PR to clarify this?

soxofaan:
> implementations could either choose to include this parameter in the canonical link href or encode it implicitly in the signed URL.

That's what I actually meant here: query parameters should be preserved in the canonical URL: visibly in the URL with a trailing `?partial=true`, or embedded in the opaque blob of the signed URL. How it's done doesn't matter, as long as the original parameter is preserved.
But that's apparently not clear from the current PR

soxofaan:
did attempt to clarify the clarification

m-mohr (approved):

PR #536, opened 2024-06-13
Document additional relation types for process metadata #531

### LINKED ISSUE #531: License for UDP?
A license field for user-defined processes might be a good idea.

m-mohr:
Closes #531

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -5265,15 +5265,17 @@ components:
             2. `version-history`: A link to a changelog and/or a list of versions of the process
             (see also the relation types `latest-version` etc.).
 
-            2. `example`: Links to examples of other processes that use this process.
+            3. `example`: Links to examples of other processes that use this process.
 
-            3. `cite-as`: For all DOIs associated with the process, the respective DOI
+            4. `cite-as`: For all DOIs associated with the process, the respective DOI
             links SHOULD be added.
 
-            4. `license`: Links to applicable license(s). The link titles should reflect the license names.
+            5. `license`: Links to applicable license(s). The link titles should reflect the license names.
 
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -5262,11 +5262,19 @@ components:
             (link to older version) and `successor-version` (link to newer version) can also be used
             to show the relation between versions.
 
+            2. `version-history`: A link to a changelog and/or a list of versions of the process
```
soxofaan, line 5265: will this duplicate  `2.` in the markdown give duplicated rendering of the "2" bullet?

soxofaan (approved):

soxofaan:
I think this PR is fine to merge as-is already

however by looking at it I'm a bit confused by this existing bit:

> latest-version: If a process has been marked as deprecated, a link SHOULD point to the preferred version of the process. The relation types predecessor-version (link to older version) and successor-version (link to newer version) can also be used to show the relation between versions.

How is this supposed to work? There is no version concept (yet) for UDPs, nor do they have a public URLs to link to .  

UDPs are only by defined by process id, so does this  documentation suggest to put the version in the the process id, e.g. `my_evi_v2`, `my_evi_v3`? (Note that dots are not allowed in process id, so this might get ugly)

openapi.yaml
m-mohr: Good point, fixed.

m-mohr:
The versioning was meant more for informative purposes, e.g. for tracability, and not necessarily lives within the API. The ID should usually be the same and old versions should not be usable anymore. At least right now. We could also integrate something on the API level, but that is future work, I think.

PR #538, opened 2024-06-20
Add Remote UDP extension #515

### LINKED ISSUE #515: Allow URLs as process namespace
To be able to load external processes (UDPs) in process graphs easily, it would be valuable to have the namespace being a URL for read-only access. The process ID stays as it is.

m-mohr:
Implements an extension to enable loading user-defined processes that are hosted externally through the process namespace into process graphs.

Closes #515

### DIFF
--- a/extensions/remote-udp/README.md
+++ b/extensions/remote-udp/README.md
@@ -22,23 +22,25 @@ This makes it possible for this extension to add additional allowed values to th
 
 This extension extends the `namespace` property of process graph nodes so that it accepts **absolute** URL with the protocols `https` (**recommended**) and `http` (discouraged). The URLs specified MUST return one of the following two options:
 
-1. A single process, compatible to the endpoint `GET /process_graphs/{process_graph_id}`.
+1. A single process, compatible\* to the endpoint `GET /process_graphs/{process_graph_id}`.
    In this case, the `id` property of the process graph node MUST be equal to the `id` of the process, 
    otherwise a `ProcessNamespaceInvalid` error is thrown
-2. A list of processes, compatible to the endpoint `GET /process_graphs`.
+2. A list of processes, compatible\* to the endpoint `GET /process_graphs`.
    In this case, the `id` property of the process graph node is used to identify the process from the list.
    If not found a `ProcessNamespaceInvalid` error is thrown
 
-### Compatibility
+\* Compatible means in this context that the requests and responses must comply to the openEO API specification with the following exceptions:
 
-Compatible means in this context that the requests and responses must comply to the openEO API specification with the following exceptions:
-
-- The `Authorization` header MUST NOT not be sent.
-- Lists of processes MUST NOT paginate and the full process description MUST be provided for each process (i.e., the recommendation to omit large properties such as `process_graph` doesn't apply).
+- User credentials / tokens that are obtained through the openEO API MUST NOT not be sent to URIs that are external to the openEO API.
+  The requirement to provide an `Authorization` header for the respective endpoints doesn't apply.
+- For a list of processes, the full process description MUST be provided for the process with the given ID within the first request.
+  This means that the recommendation to omit large roperties such as `process_graph` doesn't apply.
+  It also requires that the requester doesn't need to paginate through additional pages to find the process with the given ID.
+  Ideally, the list of processes is not paginated as otherwise the process with the given ID may move to other pages over time.
 
 ### Client Considerations
 
-Clients MUST only offer this functionality to users if the conformance class of this extension is listed in the `conformsTo` property of the `GET /` endpoint.
+If a client is conncected to a specific backend, the client MUST only offer this functionality to users if the conformance class of this extension is listed in the `conformsTo` property of the `GET /` endpoint.
 
 The protocol `http` is discouraged for URLs as web-based clients may not be able to retrieve HTTP URLs from a HTTPS context.
 For the same reason it is also RECOMMENDED to enable CORS for all URLs.
@@ -57,11 +59,11 @@ An exemplary process graph node:
 
 ```json
 {
-    "process_id": "echo",
-    "namespace": "https://hub.openeo.org/processes/echo",
-    "arguments": {
-    	"message": "Hello World"
-    },
-    "result": true
+  "process_id": "echo",
+  "namespace": "https://hub.openeo.org/processes/echo",
+  "arguments": {
+    "message": "Hello World"
+  },
+  "result": true
 }
 ```
--- a/extensions/remote-udp/README.md
+++ b/extensions/remote-udp/README.md
@@ -1,0 +1,65 @@
+# Remote UDP Extension
+
+The openEO API is a specification for interoperable cloud-based processing of large Earth observation datasets.
+
+This extension enables user to load user-defined processes that are hosted on other hosts through the process namespace into process graphs.
+
+- Version: **0.1.0**
+- Stability: **experimental**
+- Conformance class: `https://api.openeo.org/extensions/remote-udp/0.1.0`
+
+## Justification
+
+The openEO API defines the `namespace` property in a process node of a process graph as follows:
+>  The following options are predefined by the openEO API, but additional namespaces may be introduced by back-ends or in a future version of the API.
+> * `null` [...]
+> * `backend` [...]
+> * `user` [...]
+
+This makes it possible that this extension adds additional allowed values to the `namespace` property.
+
+## Specification
+
+This extension extends the `namespace` property of process graph nodes so that it accepts **absolute** URL with the protocols `https` (**recommended**) and `http` (discouraged). The URLs specified MUST either return
+
+1. a single process (compatible to the endpoint `GET /process_graphs/{process_graph_id}`)
+2. a list of processes (compatible to the endpoint `GET /process_graphs`)
+
+If a URL is provided for the `namespace` property, the `id` property of the process graph node MUST be set to the `id` of the process.
+
+### Compatibility
+
+Compatible means in this context that the requests and responses must comply to the openEO API specification with the following exceptions:
+
+- The `Authorization` header is not required and will usually not be sent.
+- Lists of processes MUST NOT paginate and the full process description MUST be provided for each process (i.e., the recommendation to omit large properties such as `process_graph` doesn't apply).
+
+### Client Considerations
+
+Clients MUST only offer this functionality to users if the conformance class of this extension is listed in the `conformsTo` property of the `GET /` endpoint.
+
+The protocol `http` is discouraged for URLs as web-based clients may not be able to retrieve HTTP URLs from a HTTPS context.
+For the same reason it is also RECOMMENDED to enable CORS for all URLs.
+
+### Error Handling
+
+The following error SHOULD be reported if the namespace can't be resolved:
+
+- Code: `ProcessNamespaceInvalid`
+- Message: `The value passed for namespace '{namespace}' in process '{process}' is invalid: {reason}`
+- HTTP Status Code: 400
+
+## Example
+
+An exemplary process graph node:
+
+```json
+{
+    "process_id": "echo",
+    "namespace": "https://hub.openeo.org/processes/echo",
+    "arguments": {
+    	"message": "Hello World"
+	},
+    "result": true
+}
+```
--- a/extensions/remote-udp/README.md
+++ b/extensions/remote-udp/README.md
@@ -2,7 +2,7 @@
 
 The openEO API is a specification for interoperable cloud-based processing of large Earth observation datasets.
 
-This extension enables user to load user-defined processes that are hosted on other hosts through the process namespace into process graphs.
+This extension enables user to load user-defined processes that are hosted external to the openEO API (e.g., GitHub or cloud storage) through the process namespace into process graphs.
 
 - Version: **0.1.0**
 - Stability: **experimental**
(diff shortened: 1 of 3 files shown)


extensions/remote-udp/README.md
```diff
@@ -0,0 +1,65 @@
+# Remote UDP Extension
+
+The openEO API is a specification for interoperable cloud-based processing of large Earth observation datasets.
+
+This extension enables user to load user-defined processes that are hosted on other hosts through the process namespace into process graphs.
```
soxofaan: "hosted on other hosts" sounds quite generic. And it's also not obvious what "other" means: other than what?

I'd propose to add a more concrete example

> (for example: external file sharing or source hosting services)

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Justification
+
+The openEO API defines the `namespace` property in a process node of a process graph as follows:
+>  The following options are predefined by the openEO API, but additional namespaces may be introduced by back-ends or in a future version of the API.
+> * `null` [...]
+> * `backend` [...]
+> * `user` [...]
+
+This makes it possible that this extension adds additional allowed values to the `namespace` property.
```
soxofaan:
```suggestion
This makes it possible for this extension to add additional allowed values to the `namespace` property.
```

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Specification
+
+This extension extends the `namespace` property of process graph nodes so that it accepts **absolute** URL with the protocols `https` (**recommended**) and `http` (discouraged). The URLs specified MUST either return
+
+1. a single process (compatible to the endpoint `GET /process_graphs/{process_graph_id}`)
+2. a list of processes (compatible to the endpoint `GET /process_graphs`)
+
+If a URL is provided for the `namespace` property, the `id` property of the process graph node MUST be set to the `id` of the process.
```
soxofaan: It took me a couple of reads to understand this statement.

> ... MUST be set to ...

I first interpreted this as instruction for the back-end to overwrite some id

> `id` of the process

when the URL provides a list or processes, there is no "the process".

Proposal:
```
1. a single process (compatible to the endpoint `GET /process_graphs/{process_graph_id}`).
   In this case, the `id` property of the process graph node MUST be equal to the `id` of the process, 
   otherwise a `ProcessNamespaceInvalid` error is thrown
3. a list of processes (compatible to the endpoint `GET /process_graphs`)
   In this case, the `id` property of the process graph node is used to identify the process from the list.
   If not found a `ProcessNamespaceInvalid` error is thrown
```

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Compatibility
+
+Compatible means in this context that the requests and responses must comply to the openEO API specification with the following exceptions:
+
+- The `Authorization` header is not required and will usually not be sent.
```
soxofaan: I think this should be "MUST not be sent" 
otherwise we risk leaking access tokens to services that should not see them

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,65 @@
...
+- Lists of processes MUST NOT paginate and the full process description MUST be provided for each process (i.e., the recommendation to omit large properties such as `process_graph` doesn't apply).
+
+### Client Considerations
+
+Clients MUST only offer this functionality to users if the conformance class of this extension is listed in the `conformsTo` property of the `GET /` endpoint.
+
+The protocol `http` is discouraged for URLs as web-based clients may not be able to retrieve HTTP URLs from a HTTPS context.
+For the same reason it is also RECOMMENDED to enable CORS for all URLs.
```
soxofaan, line 46: > RECOMMENDED to enable CORS

this is not very practical. In most cases the openEO user, client, nor backend will have control over the hosting solution.
I'm afraid web-based clients will have a hard time to resolve the URL if they want to inspect it. But I'm not sure this is a real problem as it's acctually the backend's task to  resolve the URL for evaluation

extensions/remote-udp/README.md
m-mohr:
```suggestion
- The `Authorization` header MUST NOT not be sent.
```

extensions/remote-udp/README.md
m-mohr: But if you load such a process into the Web Editor it will not work well. As such I think a recommendation is fine.

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,65 @@
...
+This makes it possible for this extension to add additional allowed values to the `namespace` property.
+
+## Specification
+
+This extension extends the `namespace` property of process graph nodes so that it accepts **absolute** URL with the protocols `https` (**recommended**) and `http` (discouraged). The URLs specified MUST either return
+
+1. a single process (compatible to the endpoint `GET /process_graphs/{process_graph_id}`)
+2. a list of processes (compatible to the endpoint `GET /process_graphs`)
+
+If a URL is provided for the `namespace` property, the `id` property of the process graph node MUST be set to the `id` of the process.
```
m-mohr:
```suggestion
1. a single process (compatible to the endpoint `GET /process_graphs/{process_graph_id}`).
   In this case, the `id` property of the process graph node MUST be equal to the `id` of the process, 
   otherwise a `ProcessNamespaceInvalid` error is thrown
3. a list of processes (compatible to the endpoint `GET /process_graphs`)
   In this case, the `id` property of the process graph node is used to identify the process from the list.
   If not found a `ProcessNamespaceInvalid` error is thrown
```

extensions/remote-udp/README.md
m-mohr:
```suggestion
This extension enables user to load user-defined processes that are hosted external to the openEO API (e.g., GitHub or cloud storage) through the process namespace into process graphs.
```

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,67 @@
...
+This makes it possible for this extension to add additional allowed values to the `namespace` property.
+
+## Specification
+
+This extension extends the `namespace` property of process graph nodes so that it accepts **absolute** URL with the protocols `https` (**recommended**) and `http` (discouraged). The URLs specified MUST either return
+
+1. a single process (compatible to the endpoint `GET /process_graphs/{process_graph_id}`).
+   In this case, the `id` property of the process graph node MUST be equal to the `id` of the process, 
+   otherwise a `ProcessNamespaceInvalid` error is thrown
+3. a list of processes (compatible to the endpoint `GET /process_graphs`)
```
m-mohr:
```suggestion
2. a list of processes (compatible to the endpoint `GET /process_graphs`)
```

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,67 @@
...
+## Specification
+
+This extension extends the `namespace` property of process graph nodes so that it accepts **absolute** URL with the protocols `https` (**recommended**) and `http` (discouraged). The URLs specified MUST either return
+
+1. a single process (compatible to the endpoint `GET /process_graphs/{process_graph_id}`).
+   In this case, the `id` property of the process graph node MUST be equal to the `id` of the process, 
+   otherwise a `ProcessNamespaceInvalid` error is thrown
+2. a list of processes (compatible to the endpoint `GET /process_graphs`)
+   In this case, the `id` property of the process graph node is used to identify the process from the list.
+   If not found a `ProcessNamespaceInvalid` error is thrown
```
m-mohr:
```suggestion
This extension extends the `namespace` property of process graph nodes so that it accepts **absolute** URL with the protocols `https` (**recommended**) and `http` (discouraged). The URLs specified MUST return one of the following two options:

1. A single process, compatible to the endpoint `GET /process_graphs/{process_graph_id}`.
   In this case, the `id` property of the process graph node MUST be equal to the `id` of the process, 
   otherwise a `ProcessNamespaceInvalid` error is thrown
2. A list of processes, compatible to the endpoint `GET /process_graphs`.
   In this case, the `id` property of the process graph node is used to identify the process from the list.
   If not found a `ProcessNamespaceInvalid` error is thrown
```

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,67 @@
...
+   otherwise a `ProcessNamespaceInvalid` error is thrown
+2. A list of processes, compatible to the endpoint `GET /process_graphs`.
+   In this case, the `id` property of the process graph node is used to identify the process from the list.
+   If not found a `ProcessNamespaceInvalid` error is thrown
+
+### Compatibility
+
+Compatible means in this context that the requests and responses must comply to the openEO API specification with the following exceptions:
+
+- The `Authorization` header MUST NOT not be sent.
```
soxofaan: I think this should state to not blindly send through the user's original openEO API bearer token to the external service.

There should still be an option for backend implementation to send other Authorization headers to secure hosting services that require some form of authentication.The details of this should not be part of this extension spec, because that would lead us to far (and need some experimentation to fully settle)

proposal:
 ```suggestion
- The user's bearer token MUST NOT be passed through automatically (e.g. as `Authorization` header) with the request to the external URL. Note that the user's bearer token should be restricted to the interaction between openEO client and openEO backend.
```

File: extensions/remote-udp/README.md [hunk 10]
```diff
@@ -0,0 +1,67 @@
...
+- Lists of processes MUST NOT paginate and the full process description MUST be provided for each process (i.e., the recommendation to omit large properties such as `process_graph` doesn't apply).
```
soxofaan: Doesn't this lead to a conflicting situation? If you list a lot of proceses *and* are forced to include the full process description, you are going to want to paginate this?

In any case, the fact that the process graph MUST be included should be move to the specification section I think, this is a hard requirement not just a compatibility  concern.

I think pagination is fine as long as the URL points to the right page directly so that the target process is included in the page. But that is kind of implicit from the current spec "`id` ... is used to identify the process from the list." Maybe what you want to note under this Compatibility section is that the backend is not expected to walk the pages (ie multiple requests) to find the process?

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Client Considerations
+
+Clients MUST only offer this functionality to users if the conformance class of this extension is listed in the `conformsTo` property of the `GET /` endpoint.
```
soxofaan: Maybe this is too much nitpicking, but can the "MUST" be relaxed to "SHOULD"? 
There are situations where you are building a process graph without knowing what the actual backend will be (e.g. because you are building generic UDPs), so there is no `GET /` for the client to check

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+An exemplary process graph node:
+
+```json
+{
+    "process_id": "echo",
+    "namespace": "https://hub.openeo.org/processes/echo",
+    "arguments": {
+    	"message": "Hello World"
+	},
```
soxofaan:
```suggestion
    },
```

:microscope:

extensions/remote-udp/README.md
m-mohr: Makes sense and I think we can make it even more clear.

> If a client is conncected to a specific backend, the client MUST

This means you can check `GET /` and in all other cases you can offer it, of course (although I believe most or all clients right now create a connection first, otherwise you may also not have a list of processes to work with).

extensions/remote-udp/README.md
m-mohr: Oh, it's part of the specification section and as such a hard requirement. It's a subsection of the specification section. It further describes the compability that is mentioned in the specification main text.

You are right that this is expressed poorly. My intention was to say that Pagination should not be made a requirement to get to the process. I tried to clarify this.

extensions/remote-udp/README.md
soxofaan: > although I believe most or all clients right now create a connection first, 

we have some internal uses of the python client where we build process graphs without a  concrete connection.

And apart from that, when building non-concrete UDPs, you should also consider that there can be a considerable difference (in time) between the moment you build the UDP and the moment you execute/use the UDP, so it doesn't make sense to be very picky about the backend capabilities at build time when they only matter at execution time.

extensions/remote-udp/README.md
m-mohr: I've tried to mitigate this by describing it better. Please re-review.

extensions/remote-udp/README.md
m-mohr: Good to know. Do you think this needs further fine-tuning?

m-mohr:
@soxofaan Updated in https://github.com/Open-EO/openeo-api/pull/538/commits/b3c7cef59a76066e29d1c81748ab4e61e64bb483  - Please check again.

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,69 @@
...
+- HTTP Status Code: 400
+
+## Example
+
+An exemplary process graph node:
+
+```json
+{
+  "process_id": "echo",
+  "namespace": "https://hub.openeo.org/processes/echo",
```
soxofaan:
```suggestion
  "namespace": "https://openeo.example.com/processes/echo",
```
(to ensure that the example url never will be valid)

extensions/remote-udp/README.md
```diff
@@ -0,0 +1,69 @@
...
+### Client Considerations
+
+If a client is conncected to a specific backend, the client MUST only offer this functionality to users if the conformance class of this extension is listed in the `conformsTo` property of the `GET /` endpoint.
+
+The protocol `http` is discouraged for URLs as web-based clients may not be able to retrieve HTTP URLs from a HTTPS context.
+For the same reason it is also RECOMMENDED to enable CORS for all URLs.
+
+### Error Handling
+
+The following error SHOULD be reported if the namespace can't be resolved:
```
soxofaan, line 50: What if the URL can be resolved, but the document does not fit? e.g. it's not JSON, it does not follow expected schema, it doesn't contain a `process_graph` property, ...?

Shouldn't this also be a `ProcessNamespaceInvalid`?
Or would that be for standard errors like `ProcessInvalid`, `ProcessGraphMissing`? Note however that these don't have placeholders like for a reason, so that could lead to poor error messages

soxofaan (commented):
some final notes, but I think the core is fine to merge as-is

extensions/remote-udp/README.md
m-mohr:
```suggestion
  "namespace": "https://openeo.example/processes/echo",
```

extensions/remote-udp/README.md
m-mohr: Yeah, I guess that should fallback to the standard errors. Generally, back-ends can also define own error messages and codes. If specific common cases occur that are not covered and need a reason, we can fine-tune this later. Please give feedback :-)

PR #549, opened 2024-11-16
Add openEO processes version #517

### LINKED ISSUE #517: How to expose which version of openeo-processes is available/targeted
I'm working on adding openeo API 1.2 and processes 2.0 to the VITO backend and the current goal is:
- https://openeo.vito.be/openeo/1.1/ (openeo API 1.1)  provides the 1.x version of openeo-processes
- https://openeo.vito.be/openeo/1.2/ (openeo API 1.2)  provides the 2.x version of openeo-processes

It would help (e.g. for user support) if we could make the linking between these versions more explicit in the metadata in some way.

Is there a way to expose the (target) openeo-processes version explicitly, e.g. globally in capabiltiies doc, or per process?

m-mohr:
Added `version` property to `GET /processes`.

Closes #517

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -615,7 +615,7 @@ paths:
                   api_version:
                     type: string
                     description: >-
-                      Version number of the openEO specification this back-end
+                      Version number of the openEO API specification this back-end
                       implements.
                     enum:
                       - 1.2.0
@@ -1660,6 +1660,17 @@ paths:
                   - processes
                   - links
                 properties:
+                  version:
+                    type: string
+                    nullable: true
+                    description: >-
+                      Version number of the openEO process specification this back-end
+                      implements.
+                      If not provided or `null` the version is either unknown or multiple versions are implemented.
+                      Clients must be able to handle multiple process versions at the same time.
+                      
+                      This is **not** the openEO API specification version.
+                    example: 1.2.0
                   processes:
                     type: array
                     items:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1666,7 +1666,7 @@ paths:
                     description: >-
                       Version number of the openEO process specification this back-end
                       targets.
-                      If not provided or `null` the version is either unknown or multiple versions are implemented.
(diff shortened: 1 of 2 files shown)


ValentinaHutter (approved):
LGTM!

openapi.yaml
```diff
@@ -1660,6 +1660,17 @@ paths:
                   - processes
                   - links
                 properties:
+                  version:
+                    type: string
+                    nullable: true
+                    description: >-
+                      Version number of the openEO process specification this back-end
+                      implements.
```
soxofaan: instead of "implements" maybe say "targets" or "uses as reference". 
I wonder if any backend fully implements the full scope of openEO processes. It would be unfortunate if no backend would set this, just because they don't implement everything or have minor deviations in experimental processes.

The main goal of this version identifier, I think, is to communicate what conventions are followed (e.g. v1 raster cubes vs v2 data cubes), or what reference is used to decide about conflicts/differences between versions.

openapi.yaml
```diff
@@ -1660,6 +1660,17 @@ paths:
...
+                      If not provided or `null` the version is either unknown or multiple versions are implemented.
```
```suggestion
                      If not provided or `null`, the version is either unknown or multiple versions are implemented.
```

openapi.yaml
```diff
@@ -1660,6 +1660,17 @@ paths:
...
+                      Clients must be able to handle multiple process versions at the same time.
```
soxofaan: I guess you mean here multiple versions of the openeo-processes specification, not multiple process versions at the same time (a backend can only define a single specification per process)

soxofaan (commented):
bit of nitpicking :smile:

openapi.yaml
m-mohr:
```suggestion
                      targets.
```

openapi.yaml
m-mohr:
```suggestion
                      Clients must be able to handle multiple process specification versions at the same time.
```

soxofaan (approved):

PR #552, opened 2024-12-13
Clarified the behaviour of `federation:backends` for `POST /validation`

m-mohr:
When implementing this in the JS client, it felt ambiguous what the property means for validation, so clarified it.

### DIFF
--- a/extensions/federation/README.md
+++ b/extensions/federation/README.md
@@ -199,7 +199,7 @@ This also covers the case where the federation supports splitting a process into
   - `federation:backends` is *not* included in the response: At least one of the back-ends support the workflow.
 - Endpoint returns errors:
   - `federation:backends` is included in the response: The listed back-ends were checked and none of the back-ends can run the workflow as is (neither splitted if supported, nor in full).
-  - `federation:backends` is *not* included in the response: None of the back-ends can run the workflow as is.
+  - `federation:backends` is *not* included in the response: the workflow could not be validated successfully by any of the back-ends or the federation component itself.
 
 ### Examples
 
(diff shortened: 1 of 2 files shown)


soxofaan:
What exactly felt ambiguous?  The error case?

It also feels a bit ambiguous to me, but I can't put my finger on it yet

m-mohr:
Generally, what the meaning of `federation:backends` in a response from `POST /validation` means.

extensions/federation/README.md
```diff
@@ -189,6 +189,18 @@ schema:
...
+
+If this property is returned through the `POST /validation` endpoint, it has the meaning as listed below.
+This also covers the case where the federation supports splitting a process into pieces so that different parts can run on different back-ends.
+
+- Endpoint returns *without* errors:
+  - `federation:backends` is included in the response: The listed back-ends support the workflow (either partially if splitting is supported, or in full).
+  - `federation:backends` is *not* included in the response: At least one of the back-ends support the workflow.
+- Endpoint returns errors:
+  - `federation:backends` is included in the response: The listed back-ends were checked and none of the back-ends can run the workflow as is (neither splitted if supported, nor in full).
+  - `federation:backends` is *not* included in the response: None of the back-ends can run the workflow as is.
```
soxofaan: I think the federation component itself (e.g. openeo-aggregator) could already raise validation issues, without any need to consult actual backends (e.g. if the whole post payload or process graph is completely nonsensical).

That would fall under "returns errors" + "federation:backends not included in response" I guess, but that might need a bit of clarification. e.g.

```suggestion
  - `federation:backends` is *not* included in the response: the workflow could not be validated successfully by any of the back-ends or the federation component itself.
```

soxofaan (approved):
I'm fine with merging this as is.

with a just a small concern to be included in this PR or a follow-up

extensions/federation/README.md
m-mohr: That sounds very reasonable.

PR #556, opened 2025-01-12
Add additional batch job timestamps

### LINKED ISSUE #542: Feature request: add `started`, ... to job metadata
job metadata at `GET /jobs/{job_id}` currently lists these timestamps:
- `created` (required): Date and time of creation
- `updated` (optional): Date and time of the last status change

This is a feature request to add 
- `started` (optional): date and time when the job was started (`POST /jobs/{job_id}/results`)
- `stopped` (optional): date and time when job stopped running (because of reaching status finished/error/canceled)

Context: we are handling some  larger openEO use cases where a significant number of jobs has to be managed. We noticed that the "created" timestamp is not always a very informing aspect, while a "started" timestamp would be more relevant. For example because the jobs are created in bulk in advance, while they are started over a longer period, possibly hours or days after creation.

m-mohr:
Implements #542

Reasoning for the fields based on the discussion in #542 and the existing fields (created, updated, expires):

- created: exists, the time when the batch job was created
- updated: exists, the time of the last status change
- queued: new, the time when the job was queued, to be able to check how long jobs are queued (with started)
- started: new, the time when the job started processing (status changed to running), to be able to check how long processing takes (with updated)
- expires: partially new, the time when the assets get deleted (or if in the past, the time when the assets got deleted) - a field with this description existed in the STAC metadata, added to the job metadata. Would be great if that gets filled if results get deletes so that clients can flag these jobs.
- ended: NOT added, it would always be the same value as updated according to the definition of the updated field.

All properties are optional except for created.

created, updated and expires are somewhat aligned in their meaning with STAC. queued and started don't exist in STAC.


clausmichele (approved):

clausmichele:
An useful addition to track the time a batch job takes without workarounds, thanks! By the way, to add those field to the STAC results we would have to use this extension right? https://github.com/stac-extensions/timestamps/

m-mohr:
@clausmichele Only for expires, all others are not defined in that extension. Due to slightly different meaning, it could also be omitted.

openapi.yaml
```diff
@@ -5660,16 +5665,39 @@ components:
...
+        Date and time of the last status change,
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+        If the status is `error`, `canceled` or `finished`,
+        this is the time when the job has ended.
       example: '2017-01-01T09:36:18Z'
+    queued:
+      type: string
+      format: date-time
+      description: >-
+        Date and time of queueing the batch job (i.e. when the status 'queued' was set),
```
SergeCroise:
```suggestion
        Date and time of queueing the batch job (i.e., when the status 'queued' was set),
```

openapi.yaml
```diff
@@ -5660,16 +5665,39 @@ components:
...
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+      example: '2017-01-01T09:34:00Z'
+    started:
+      type: string
+      format: date-time
+      description: >-
+        Date and time when the batch job started processing (i.e. when the status 'running' was set),
```
SergeCroise:
```suggestion
        Date and time when the batch job started processing (i.e., when the status 'running' was set),
```

openapi.yaml
```diff
@@ -5660,16 +5665,39 @@ components:
...
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
       example: '2017-01-01T09:32:12Z'
     updated:
       type: string
       format: date-time
       description: >-
-        Date and time of the last status change, formatted as a [RFC
-        3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+        Date and time of the last status change,
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```
SergeCroise:
```suggestion
        formatted as an [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```

openapi.yaml
```diff
@@ -5660,16 +5665,39 @@ components:
       type: string
       format: date-time
       description: >-
-        Date and time of creation, formatted as a [RFC
-        3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+        Date and time of creation (for batch jobs: the status 'created' was set),
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```
SergeCroise:
```suggestion
        formatted as an [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```

openapi.yaml
```diff
@@ -5660,16 +5665,39 @@ components:
...
+      description: >-
+        Date and time when the batch job started processing (i.e. when the status 'running' was set),
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+      example: '2017-01-01T09:36:18Z'
+    expires:
+      type: string
+      format: date-time
+      description: >-
+        Time until which the assets are accessible, in UTC. Formatted as
+        a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```
SergeCroise:
```suggestion
        an [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```

openapi.yaml
```diff
@@ -5660,16 +5665,39 @@ components:
...
+      format: date-time
+      description: >-
+        Date and time when the batch job started processing (i.e., when the status 'running' was set),
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+      example: '2017-01-01T09:36:18Z'
+    expires:
+      type: string
+      format: date-time
+      description: >-
+        Time until which the assets are accessible, in UTC. Formatted as
```
```suggestion
        Time until which the assets (e.g., batch job results) are available for download, in UTC. Formatted as
```

soxofaan (commented):
minor suggestion

openapi.yaml
m-mohr: That might be a bit misleading. 
expires can be used for when the signed URLs expire and also for when the data is deleted, I think. That's why I chose accessible. The addition in brackets is fine though.

```suggestion
        Time until which the assets (e.g., batch job results) are accessible, in UTC. Formatted as
```

Maybe we should actually split into expires and unpublished, i.e.

- expires: signed URL expiry (if relevant here?) but in STAC it's that...
- unpublished: when the data gets deleted

soxofaan: ok good point. 
But indeed, we should separate the expiry of signed URLs (which can be refreshed easily) and the expiry for download availability in general (which is final and non-recoverable).

I think the general availability of results should be a global property on  `GET /jobs/{job_id}` or `GET /jobs/{job_id}/results`, while the signed URL expiry should be closer to the links/URLs themselves (e.g. different signed URLs within the same resource might have different expiry for some advanced reason)

m-mohr: Okay, I've updated the PR accordingly:

- Clarified usage of expires, removed from the batch job info endoint
- Added unpublished to batch job info endpoint and batch job results
- Added all the timestamps also to the Collection-typed result response, previously it was only defined in the Item-typed result response

openapi.yaml
```diff
@@ -5660,16 +5681,49 @@ components:
...
+        formatted as an [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+        If the status is `error`, `canceled` or `finished`,
+        this is the time when the job has ended.
+      example: '2017-01-01T09:54:18Z'
+    queued:
+      type: string
+      format: date-time
+      description: >-
+        Date and time of queueing the batch job (i.e., when the status 'queued' was set),
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```
SergeCroise:
```suggestion
        formatted as an [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```

openapi.yaml
```diff
@@ -5660,16 +5681,49 @@ components:
...
+      example: '2017-01-01T09:34:00Z'
+    started:
+      type: string
+      format: date-time
+      description: >-
+        Date and time when the batch job started processing (i.e., when the status 'running' was set),
+        formatted as a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```
SergeCroise:
```suggestion
        formatted as an [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
```

m-mohr:
@soxofaan I think we can also merge this, if VITO is happy with it.

openapi.yaml
```diff
@@ -5666,16 +5687,49 @@ components:
...
+      type: string
+      format: date-time
+      description: >-
+        Time in UTC until which the assets and this document are accessible via the signed URL
+        that is provided with the relation type `canonical` in the links.
+        Formatted as an [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339.html) date-time.
+        After expiration, a document with new signed URLs can be retrieved through an authenticated
+        request to this endpoint.
+      example: '2017-02-01T09:54:18Z'
+    unpublished:
```
soxofaan, line 5726: I'm a bit hesitant about "unpublished". That word could suggest that the results still should be published, or can be recovered/republished. I think this should sound more irrevocable, like "removed", "erased" or "purged"

m-mohr: The API doesn't describe whether that's permanent or not, it just tells the user that the data is currently not available.
Both cases are valid. In your case it's permanent, but it might not be in other cases.

PR #560, opened 2025-03-05
Federation extension: nullable last_successful_check

soxofaan:
when back-end was never observed to be available

### DIFF
--- a/extensions/federation/README.md
+++ b/extensions/federation/README.md
@@ -59,10 +59,13 @@ schema:
             format: date-time
             description: The time at which the status of the back-end was checked last.
           last_successful_check:
-            type: string
-            format: date-time
+            type:
+              - type: string
+                format: date-time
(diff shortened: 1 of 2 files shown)


soxofaan:
not sure if this minor tweak deserves a changelog entry

extensions/federation/README.md
```diff
@@ -59,10 +59,13 @@ schema:
             format: date-time
             description: The time at which the status of the back-end was checked last.
           last_successful_check:
-            type: string
-            format: date-time
+            type:
+              - type: string
+                format: date-time
+              - type: null
```
m-mohr: We are still on OpenAPI 3.0...
```suggestion
            type: string
            format: date-time
            nullable: true
```

m-mohr:
Thanks. A changelog entry would be great, I think.

PR #565, opened 2025-06-30
PR #564 finetuning of back-end "availability" terminology

### LINKED ISSUE #562: Clarifications for the Federation extension

soxofaan:
This is an auxiliary PR for #564 . To be merged or cherry-picked from

Main goal is to eliminate/reduce  "permanent"/"temporary" terminology which might make cause a mismatch in expectations as I tried to argue in https://github.com/Open-EO/openeo-api/pull/564#issuecomment-2909544291 

contents:
- explicitly mention  properties `federation:backends` and `federation:missing` in the descriptions
- eliminate the word "permanently" (used once for `federation:backends`), which was redundant in my opinion
- put `list of resources` in bold to (hopefully) make it more clear that `federation:backends` is for individual resources and `federation:missing` is for compiled listings
- reduce usage of "temporary"  and replace it with a neutral "not available". I kept it when used in combination "temporarily or unintentionally" where it has value.
- small typo fix of "the a combination"

### DIFF
--- a/extensions/federation/README.md
+++ b/extensions/federation/README.md
@@ -108,8 +108,10 @@ schema:
 
 ## Resources supported only by a subset of back-ends
 
-Every discoverable resource that is defined as an object and allows to contain additional properties, can list the subset of back-ends that permanently support or host the exposed resource/functionality.
-Examples of where this could apply to (**not** comprehensive):
+Discoverable resources can explicitly list the subset of back-ends that support or host the exposed resource or functionality with the property `federation:backends`.
+
+Schema-wise, this only applies to resources that are defined as an object and allow to contain additional properties.
+For example (**not** comprehensive):
 
 - `GET /collections`
 - `GET /collections/{id}`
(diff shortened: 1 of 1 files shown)


m-mohr:
Thanks, that sounds good.

Issue #338, opened 2020-10-21
Bearer token not JWT conform


aljacob:
Today while implementing authorization I realized that our way of defining the token we send as access token is not JWT conform and can create all kinds of issues when relying on external libraries or frameworks for parsing the token.
Can we remove the /oidc/provider/ before the token, so that we just send Bearer token instead of Bearer /oidc/provider/token. If we need for some reason the information on which auth method and which provider has been chosen I would rather just add this as additional custom header fields instead of encoding them in the Bearer String.

m-mohr:
The Bearer Token has the actual token after the prefix. The API doesn't mandate JWT to be used due to different authentication procedures. See the API description: 

> The Bearer Token MUST consist of the authentication method, a provider ID (if available) and the token itself. All separated by a forward slash `/`. Examples (replace `TOKEN` with the actual access token): (1) Basic authentication (no provider ID available): `basic//TOKEN` (2) OpenID Connect (provider ID is `ms`): `oidc/ms/TOKEN`. For OpenID Connect, the provider ID corresponds to the value specified for `id` for each provider in `GET /credentials/oidc`.

The TOKEN above is your JWT, so you need to split the Bearer token at the `/` char and the third part is your JWT you can then parse.

We can't change this easily, it would be a breaking change.

soxofaan:
I agree with @m-mohr , the bearer token that is sent from openEO client ot openEO backend should be handled as a opaque blob. I don't think there is a standard that requires it to be a JWT token.  

Also, I noticed lately that JWT gets some bad press from time to time, so we shouldn't bet too big on it. It is not unlikely that some providers switch to another solution in the future.

aljacob:
maybe I expressed myself poorly. The token must not be expected to be a JWT token in general. I agree with this. But if somebody relies on tools that are using JWT tokens, then the validators in those tools will fail by default with the current definition, as you are basically manipulating the token that has been send to you, so when you send it back for validation to the backend from which you got it, it does not get the same thing as it expects and hence validation fails.

m-mohr:
What do you expect us to do now? If you want a change here you need to write a proposal to the PSC. I still think you can solve this in implementation though.

soxofaan:
>  you are basically manipulating the token that has been send to you, so when you send it back for validation to the backend from which you got it, it does not get the same thing as it expects and hence validation fails.

[...]

PR #567, opened 2025-08-22
Migrate from openEO-specific tokens to JWT #558

### LINKED ISSUE #558: Eliminate custom bearer token format
In the openEO API spec we currently require bearer tokens of the format `oidc/{provider_id}/{token}` or `basic//{token}`:

https://github.com/Open-EO/openeo-api/blob/c5a45b4647b06e313a4f099e9119bfa3cca5c6a3/openapi.yaml#L6714-L6721

My view is a bit limited and openEO biased, but it seems there is growing convention to use a JWT access token as bearer token in the `Authorization` header, and authentication/authorization related tooling (e.g. security/authorization gateways) builds on that assumption.
The custom prefix format we currently use of course conflicts with that.
I wonder if we can figure something out to make the openEO API more compatible with that assumption/convention.

e.g.:
- if basic auth is disabled and  there is only one OIDC provider active: it's allowed to omit the prefix
- split off the method and OIDC provider to a different request header

This would clearly be a breaking change, so not something in scope of the standard openEO API spec v1.x. 
But maybe it could be an optional extension in some way.

(This came up again in another EOEPCA meeting, so I though it would be good to kick off some discussion here or at least have a pointer to that aspect)

m-mohr:
Closes #558 and #338

Should be fully backward compatible as specified.


CHANGELOG.md
```diff
@@ -15,6 +15,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `version` property to `GET /processes` [#517](https://github.com/Open-EO/openeo-api/issues/517)
 - Added `queued`, `started` and `unpublished` to the batch job metadata and the corresponding STAC results [#542](https://github.com/Open-EO/openeo-api/issues/542)
 - Added all the batch job timestamps (including the new timestamps above) to the Collection type of batch job results
+- Support for JWT-based Bearer tokens [#558](https://github.com/Open-EO/openeo-api/issues/558)
```
soxofaan: The current tokens are  already JWT "based" (prefix + JWT) in practice (on most deployments), so this might be a bit confusing.
Suggestion to clarify that a bit:

```suggestion
- Support for standard JSON Web Tokens (JWT) being used as Bearer tokens [#558](https://github.com/Open-EO/openeo-api/issues/558)
```

CHANGELOG.md
```diff
@@ -15,6 +15,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Deprecated
+
+- Deprecated the openEO-specific Bearer token format (authentication mechanism/provider id/access token) [#558](https://github.com/Open-EO/openeo-api/issues/558)
+
+### Changed
+
+- Migrate from openEO-specific tokens to JWT [#558](https://github.com/Open-EO/openeo-api/issues/558)
```
soxofaan: What is migrated here? This is just about deprecating the old way and recommending a new way, right?

openapi.yaml
```diff
@@ -6808,14 +6789,46 @@ components:
...
+             For OpenID Connect, the provider ID corresponds to the value
+             specified for `id` for each provider in `GET /credentials/oidc`.
+
+           All servers must accept this method for backward compatibility
+           until version 2.0 of the specification.
+
+          The access tokens provided by the identity provider MUST NOT include
+          the prefix that includes the authentication method and provider ID.
+          The Bearer Token sent to the server MUST be prefixed with `basic//`.
+          This means that the clients have to prepend the prefix.
```
soxofaan: I understand that this paragraph is moved over from the basic auth section above, but putting it here makes things more confusing I think.

e.g. 

> access tokens provided by the identity provider MUST NOT include the prefix

In the most common case (external OIDC provider), the reader has no control over that, while the current wording seems to suggest that. I know that it's related to the `/credentials/basic` endpoint, but then I would clarify that there (as original).

> ... MUST be prefixed with `basic//`. ...

This is probably a copy-paste issue, but the text also applies to OIDC, not just basic auth.


It also seems the paragraph is indented with one character less than the precious paragraph, which might cause some rendering issues?

CHANGELOG.md
m-mohr: Yes

openapi.yaml
```diff
@@ -6808,14 +6789,46 @@ components:
...
+           - OpenID Connect (provider ID is `ms`): `oidc/ms/TOKEN`.
+             For OpenID Connect, the provider ID corresponds to the value
+             specified for `id` for each provider in `GET /credentials/oidc`.
+
+           All servers must accept this method for backward compatibility
+           until version 2.0 of the specification.
+
+          The access tokens provided by the identity provider MUST NOT include
+          the prefix that includes the authentication method and provider ID.
+          The Bearer Token sent to the server MUST be prefixed with `basic//`.
```
m-mohr:
```suggestion
          The Bearer Token sent to the server MUST have the prefix, e.g. `basic//` for Basic authentication.
```

openapi.yaml
m-mohr: Would https://github.com/Open-EO/openeo-api/pull/567/files#r2310265183 be enough to solve the issue?

With the for example, all other parts are generic again and actually do apply to OIDC and Basic.

soxofaan:
One conflict that I'm still pondering on:

The basic auth option does not play nicely with JWT I think. The point of basic auth is to have a very simple, low overhead auth mechanism (a single `/credentials/basic` endpoint to produce tokens and the generic `Authorization` header for consumption).  This is ideal for sandboxed testing/benchmark use cases, where one doesn't want to spend too much effort on a full auth stack.   

To support JWT and the related mechanisms however you need to do quite a bit more: you have to set up an internal issuer service, possibly manage secrets (for signing), support multiple extra endpoints (`.well-known/openid-configuration`, `/token`, `/userinfo`, ...)

I'm not sure we should aim for removal of the openEO token format, as it will make basic auth practically unusable.

I would aim for something like:
- OIDC:
    - recommend standard JWT
    - deprecate openEO token format, possibly removing support in later openEO API version
- basic auth:
    - keep openEO token format
    - allow JWT usage, but nobody is probably going to do that

m-mohr:
Is that really the case? Can't you just create a JWT using one of the JWT libraries? JWT is completely independant of OIDC, you certainly don't need the extra endpoints. https://jwt.io

openapi.yaml
soxofaan: > Would https://github.com/Open-EO/openeo-api/pull/567/files#r2310265183 be enough to solve the issue?

That improves the situation of my second point indeed

But  I still think that the following sentence (which talks about an _access_ token), might be confusing in the description of the _bearer_ token. 

> The access tokens provided by the identity provider MUST NOT include
> the prefix that includes the authentication method and provider ID.

It's also only relevant as `/credentials/basic` documentation, but has no relevance to OIDC

openapi.yaml
```diff
@@ -6808,14 +6789,46 @@ components:
...
+
+           If this method is supported by the openEO implementation,
+           the server MUST provide the conformance class given above.
+
+        2. **openEO Tokens - DEPRECATED**
+
+           - Conformance class: *None*
+
+           The Bearer Token is constructed from the authentication method, a
+           provider ID (if available) and the token itself. All separated by a
```
```suggestion
           provider ID (if available) and the access token. All separated by a
```

openapi.yaml
```diff
@@ -6808,14 +6789,46 @@ components:
...
-        provider in `GET /credentials/oidc`.
+      bearerFormat: JWT or openEO
+      description: >-
+        Bearer tokens can be provided in two different ways:
+
+        1. **JSON Web Token (JWT) - RECOMMENDED**
+
+           - Conformance class: `https://api.openeo.org/1.2.0/authentication/jwt`
+           
+           The Bearer token is a [JWT](https://datatracker.ietf.org/doc/html/rfc7519)
```
```suggestion
           The Bearer token is an access token in [JWT](https://datatracker.ietf.org/doc/html/rfc7519) format
```

openapi.yaml
```diff
@@ -6808,14 +6789,46 @@ components:
...
-        The Bearer Token MUST consist of the authentication method, a provider
-        ID (if available) and the token itself. All separated by a forward slash
-        `/`. Examples (replace `TOKEN` with the actual access token): (1) Basic
-        authentication (no provider ID available): `basic//TOKEN` (2) OpenID
-        Connect (provider ID is `ms`): `oidc/ms/TOKEN`. For OpenID Connect, the
-        provider ID corresponds to the value specified for `id` for each
-        provider in `GET /credentials/oidc`.
+      bearerFormat: JWT or openEO
+      description: >-
+        Bearer tokens can be provided in two different ways:
```
```suggestion
        A bearer token is based on an access token and can be provided in two different formats:
```

soxofaan (commented):
some additional suggestions to further clarify the subtle differences between access and bearer tokens

soxofaan:
> Is that really the case? Can't you just create a JWT using one of the JWT libraries? JWT is completely independant of OIDC, you certainly don't need the extra endpoints.

Creating JWT tokens is indeed not a problem. The problem is that the backend must decode and verify that the tokens effectively come from the trusted OIDC issuer. So as far as I currently understand, you need one of:
- get some public pub key of the OIDC issuer (a .well-known mechanism) to use in signature verification
- send the token to the `/userinfo` endpoint of the issuer (which also involves discovery from .well-known/...) to let the issuer do the verification

And I even think that the second option is necessary to obtain extra info that is not in the token itself (e.g. user roles or that virtual org membership stuff from EGI)

Using `/userinfo` to do the verification and enrichment  is what we currently do in the VITO implementation FYI

m-mohr:
Well, but if the backend detects that it's from basic (e.g. by including that into the JWT), it can just skip all OIDC stuff and just check whether that token is valid. Whatever that means for the Basic JWT tokens, is up to the provider to decide. I don't really see a big difference to before, except that it's now JWT encoded.

I'll go through the proposed textual changes after next week. :-)

openapi.yaml
m-mohr: I'm not sure I understand, I think this is relevant for all authentication methods. In any case, it shall not include the openEO proprietary prefix.

CHANGELOG.md
m-mohr:
```suggestion
- Migrate from openEO-specific tokens to JWT, i.e. deprecating the openEO-specific format in favor of JWT [#558](https://github.com/Open-EO/openeo-api/issues/558)
```

openapi.yaml
m-mohr: Why is it "based on" an access token?

openapi.yaml
m-mohr: I'm not sure I fully agree with the access token terminology in the proposals. Does it always need to be an access token outside of OIDC?

In any case, it should be consistent in the PR, which it is not fully, but I'm not sure whether we should always refer to access token. In the Basic case it could potentially something different, and in other (future) auth mechanisms, too?!

m-mohr: Actually, nevermind. We use access token as property name in basic.

openapi.yaml
m-mohr:
```suggestion
        A Bearer token can be provided in two different formats:
```

openapi.yaml
soxofaan: good point actually, "access token" is OIDC-specific and generalizing to the basic auth case might trigger confusion

openapi.yaml
```diff
@@ -6808,14 +6789,46 @@ components:
...
+      bearerFormat: JWT or openEO
+      description: >-
+        A Bearer token can be provided in two different formats:
+
+        1. **JSON Web Token (JWT) - RECOMMENDED**
+
+           - Conformance class: `https://api.openeo.org/1.2.0/authentication/jwt`
+           
+           The Bearer token is an access token in [JWT](https://datatracker.ietf.org/doc/html/rfc7519) format
+           as defined in RFC 7519. For openEO, it MUST include the issuer in the
```
soxofaan, line 6803: "issuer" is an OIDC-specific concept, so this "MUST" would exclude using JWT for basic auth.

soxofaan: a solution would be to specify that the openeo endpoint "https://{backend}/credentials/basic" must be used as issuer for basic auth, or something like that

openapi.yaml
```diff
@@ -6808,14 +6789,46 @@ components:
...
+           `iss` claim although being optional in RFC 7519.
+
+           If this method is supported by the openEO implementation,
+           the server MUST provide the conformance class given above.
```
soxofaan: I would also add that an openeo client can only use JWT if the backend lists the conformance class

openapi.yaml
soxofaan: What I'm trying to get across:
This whole description is about the responsibilities of an **openEO client** to properly provide the bearer token for an **openEO backend** to understand.

This is the only exception:
>  The access tokens provided by the identity provider MUST NOT include the prefix that includes the authentication method and provider ID.

That sentence discusses the responsibility of the **identity provider** on how to provide the token to the **openEO client**. But:
- in case of OIDC, it makes little sense to talk about that, because the how of that is fully specified in the OIDC spec, and it's completely outside the scope of the openEO API, neither the openEO client nor backend have any influence on that. I'm just afraid that discussing this in the description here (with "MUST NOT" language) might confuse the reader about how things work and interact, because it gives the wrong impression that this aspect is part of the openEO API, or  influenceable by the openEO client or backend.
- it is only relevant for the response of `/credentials/basic`, which is the only place in the openEO API were an **openEO backend** provides a token to an **openEO client**. 

So that's why I think that this sentence is kept in its original location (the description of `/credentials/basic`)  instead of in this generic place.

openapi.yaml
m-mohr: I was thinking about that, but didn't want to prescribe a specific value.

openapi.yaml
m-mohr: Uhm, I don't understand the difference to the status quo... 🤔

openapi.yaml
m-mohr: I changed it slightly to say "do not" instead of "MUST NOT". It makes sense to not require IP behaviour here, but it is also needed to have this introductory section for the following two sentences that describe the client behavior.

The basic endpoint already states that the "basic//" prefix should not be included and OIDC is well-defined anyway.

So I think that hopefully solves it.

openapi.yaml
m-mohr: Added:

> If the concept of an issuer doesn not exist in an authentication method (e.g. in HTTP Basic), implementations could use the endpoint for Basic Authentication as the issuer, for example.

openapi.yaml
soxofaan: I guess I was confused by what the sentence actually states. It only talks about the responsibility of the server, but the client must also check for the presence of the conformance class before using JWT.

Proposal:

> openEO backend implementations MUST signal their support for JWT by listing the conformance class in the capabilities document.
> Likewise, openEO clients SHOULD only use JWT when the openEO backend list the conformance class.

openapi.yaml
```diff
@@ -6786,14 +6767,48 @@ components:
...
+
+           - Basic authentication (no provider ID available): `basic//TOKEN`
+           - OpenID Connect (provider ID is `ms`): `oidc/ms/TOKEN`.
+             For OpenID Connect, the provider ID corresponds to the value
+             specified for `id` for each provider in `GET /credentials/oidc`.
+
+           All servers must accept this method for backward compatibility
+           until version 2.0 of the specification.
+
+          The access tokens provided by the identity provider do not include
```
soxofaan: I think there is something off identation-wise here. Selectively picking particular lines to highlight:

```
        2. **openEO Tokens - DEPRECATED**
           All servers must accept this method for backward compatibility
          The access tokens provided by the identity provider do not include
```

soxofaan (commented):
did another quick review round

openapi.yaml
m-mohr: Done, thanks.

openapi.yaml
m-mohr: Thanks for spotting this, fixed.

m-mohr:
@soxofaan I think all comments have been resolved. Can you re-review, please?

openapi.yaml
```diff
@@ -6810,14 +6791,49 @@ components:
...
+        A Bearer token can be provided in two different formats:
+
+        1. **JSON Web Token (JWT) - RECOMMENDED**
+
+           - Conformance class: `https://api.openeo.org/1.2.0/authentication/jwt`
+           
+           The Bearer token is an access token in [JWT](https://datatracker.ietf.org/doc/html/rfc7519) format
+           as defined in RFC 7519. For openEO, it MUST include the issuer in the
+           `iss` claim although being optional in RFC 7519.
+           If the concept of an issuer doesn not exist in an authentication method (e.g. in HTTP Basic),
```
```suggestion
           If the concept of an issuer does not exist in an authentication method (e.g. in HTTP Basic),
```

openapi.yaml
```diff
@@ -6810,14 +6791,49 @@ components:
...
+           forward slash `/`.
+
+           Examples (replace `TOKEN` with the actual access token):
+
+           - Basic authentication (no provider ID available): `basic//TOKEN`
+           - OpenID Connect (provider ID is `ms`): `oidc/ms/TOKEN`.
+             For OpenID Connect, the provider ID corresponds to the value
+             specified for `id` for each provider in `GET /credentials/oidc`.
+
+           All servers must accept this method for backward compatibility
```
```suggestion
           All openEO backends MUST accept this method for backward compatibility
```

openapi.yaml
```diff
@@ -6810,14 +6791,49 @@ components:
...
+           until version 2.0 of the specification.
```
soxofaan, line 6828: Is it intentional to say "until" instead of, e.g., "at least until"?
Is the implication here that the classic openEO token-style would become invalid starting with 2.0?

soxofaan (commented):
some more suggestions/minor questions

openapi.yaml
m-mohr: Yes, the intention is to deprecate in v1.3 and remove in v2.0.

openapi.yaml
```diff
@@ -6810,14 +6791,49 @@ components:
...
+
+           The access tokens provided by the identity provider do not include
+           the prefix that includes the authentication method and provider ID.
+           The Bearer Token sent to the server MUST have the prefix, e.g. `basic//` for Basic authentication.
```
m-mohr:
```suggestion
           The Bearer Token sent to the openEO backend MUST have the prefix, e.g. `basic//` for Basic authentication.
```

m-mohr:
Thanks, updated.

soxofaan (approved):
👍  ready to merge I think

PR #572, opened 2025-09-15
Clarifies the terms "end"/"result" nodes and "root"/"child" process graphs

### LINKED ISSUE #547: Confusing and contradicting coverage of  "result" flag and multiple end/final nodes in process graphs
Some excerpts of the spec (current draft branch): [L298](https://github.com/Open-EO/openeo-api/blob/162952e62012a667172bcf9ec222712041f6a828/openapi.yaml#L298) [L326](https://github.com/Open-EO/openeo-api/blob/162952e62012a667172bcf9ec222712041f6a828/openapi.yaml#L326) [L474](https://github.com/Open-EO/openeo-api/blob/162952e62012a667172bcf9ec222712041f6a828/openapi.yaml#L474) 

> Technically, a process graph is defined to be a graph of connected processes with exactly one node returning the final result 
> ...
> .. multiple end nodes are possible

"exactly one node returning the final result" is in conflict with the statement that multiple end nodes are possible, as discussed here as well:
- https://github.com/Open-EO/openeo-api/issues/427

>  One of the nodes in a map of processes (the final one) MUST have the `result` flag set to `true`, all the other nodes can omit it as the default value is `false`. 

It's unclear if the "One of the nodes" here means "exactly one" or "at least one". I guess the intention is  "exactly one", but the current wording still allows the (wrong) interpretation that multiple nodes can have `result=true`.
Also, as discussed at https://github.com/Open-EO/openeo-processes/issues/279#issuecomment-937834642, the presence of a result node is actually only necessary for "child" processes, so the spec could also specify "at most one".

Another thing to nitpick about in that sentence is "the final one":
- seems to suggest there is a single  "final" node, which is not true if you heave multiple "end" nodes in your graph
- it suggests "final node" and "result node" are synonyms, but I don't think we should establish such an equivalence: you can have multiple "final" nodes, but only one  "result" node. Even the spec itself suggest that:
    > Please be aware that the result node (`result` set to `true`) is not necessarily the last node that is executed. The author of the process graph may choose to set a non-end node to the result node!

> Having such a node is important as multiple end nodes are possible, but in most use cases it is important to exactly specify the return value to be used by other processes. Each child process graph must also specify a result node similar to the "main" process graph. 

First, a nitpicking note: the first sentence here introduces "end node". So at this point, the spec is mixing usage of "result node", "final node" and "end node", in some places it vaguely suggest these are equivalent, while in other places it contradicts this.

Regardless, the message of these two sentences is a bit unclear to me. The first sentence seems to start talking about the top-level graph (where multiple end nodes are a valid use case), but then with the "used by other processes" it seems to switch over to child process graphs. But then actually the next sentence talks about "child process graphs" explaining the need for a result node based on the "main" process graph. While it's actually the other way around: child process graphs need a single result node, while the "main" process graph doesn't.

m-mohr:
Clarified the relation between "result nodes" and "end nodes", the usage of the `result` flag, and the relation between the outermost ("root") and "child" process graphs

Closes #547

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -323,9 +323,13 @@ info:
     ```
     A process node MUST always contain key-value-pairs named `process_id` and `arguments`. It MAY contain a `description`.
 
-    One of the nodes in a map of processes (the final one) MUST have the `result` flag set to `true`, all the other nodes can omit it as the default value is `false`. Having such a node is important as multiple end nodes are possible, but in most use cases it is important to exactly specify the return value to be used by other processes. Each child process graph must also specify a result node similar to the "main" process graph.
-
     `process_id` MUST be a valid process ID in the `namespace` given. Clients SHOULD warn the user if a user-defined process is added with the same identifier as one of the predefined process.
+
+    In the following, "end node" (also known as "leaf node") defines a node that is not referenced by any other node in the same process graph. The "root process graph" is the outermost process graph, which is not part of any other process graph.
+
+    A "result node" is a node that defines the return value of the process graph and has the `result` flag set to `true`. Exactly one of the nodes in a map of processes MUST have the `result` flag set to `true`, all the other nodes can omit it as the default value is `false`. Each child process graph MUST specify its own result node. The root process graph MUST also specify a result node although not strictly needed in all use cases.
+    
+    Having a result node is important as multiple end nodes are possible and in many use cases it is important to identify a specific return value for the process graph that is passed to other processes. The result node is not necessarily an end node.
 
     ### Arguments
 
@@ -437,7 +441,7 @@ info:
 
     ### Full example for an EVI computation
 
-    Deriving minimum EVI (Enhanced Vegetation Index) measurements over pixel time series of Sentinel-2 imagery. The main process graph in blue, child process graphs in yellow:
+    Deriving minimum EVI (Enhanced Vegetation Index) measurements over pixel time series of Sentinel-2 imagery. The root process graph in blue, child process graphs in yellow:
 
     ![Graph with processing instructions](assets/pg-evi-example.png)
 
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -327,7 +327,7 @@ info:
 
     In the following, "end node" (also known as "leaf node") defines a node that is not referenced by any other node in the same process graph. The "root process graph" is the outermost process graph, which is not part of any other process graph.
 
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -323,10 +323,14 @@ info:
     ```
     A process node MUST always contain key-value-pairs named `process_id` and `arguments`. It MAY contain a `description`.
 
-    One of the nodes in a map of processes (the final one) MUST have the `result` flag set to `true`, all the other nodes can omit it as the default value is `false`. Having such a node is important as multiple end nodes are possible, but in most use cases it is important to exactly specify the return value to be used by other processes. Each child process graph must also specify a result node similar to the "main" process graph.
-
     `process_id` MUST be a valid process ID in the `namespace` given. Clients SHOULD warn the user if a user-defined process is added with the same identifier as one of the predefined process.
 
+    In the following, "end node" (also known as "leaf node") defines a node that is not referenced by any other node in the same process graph. The "root process graph" is the outermost process graph, which is not part of any other process graph.
```
soxofaan, line 328: > "end node" (also known as "leaf node")

bike-shedding here: if I would have to pick, I'd prefer "leaf node" as main concept, as it feels more self-explanatory than "end node"

openapi.yaml
```diff
@@ -323,10 +323,14 @@ info:
...
+
+    A "result node" is a node that defines the return value of the process graph and has the `result` flag set to `true`. Exactly one of the nodes in a map of processes MUST have the `result` flag set to `true`, all the other nodes can omit it as the default value is `false`. Each child process graph MUST specify its own result node. The root process graph MUST also specify a result node although not strictly needed in all use cases.
```
soxofaan: Although it is mentioned lower alread, I'd put this a bit earlier in the definition:
```suggestion
    A "result node" is a node (often an end node, but that's not required) that defines the return value of the process graph and has the `result` flag set to `true`. Exactly one of the nodes in a map of processes MUST have the `result` flag set to `true`, all the other nodes can omit it as the default value is `false`. Each child process graph MUST specify its own result node. The root process graph MUST also specify a result node although not strictly needed in all use cases.
```

soxofaan (commented):
Some nit-picking, but I would be fine with merging this as is

openapi.yaml
m-mohr: I've just tested with two non-openEO persons and they both seemed to better understand end node. So I'm not sure what's actually more self-explanatory and would stick with it for now.

PR #581, opened 2025-12-04
`GET /service_types`: Add `specification` property

### LINKED ISSUE #580: Identify the type of web service
Currently implementing the QGIS plugin, and we can't really identify easily which services are actually which type. For example, GEE uses `type: xyz` for XYZ, but that's not standardized. I think we should recommend some specific values in the spec for certain service types.

For example:
- ogcapi-maps
- ogcapi-tiles
- ogcapi-features
- wms
- wmts
- xyz

To allow some freedom, we could also just recommand to let the type start with this and then people could add something else.
Alternatively, we leave the type unspecified but add an identifier for the web service type to the GET /service_types response Potentially reuse the naming of the STAC web-map-links extension.

m-mohr:
`GET /service_types`: Added `specification` property to identify the actual specification the exposed services implement.

Closes #580


m-mohr:
Might be interesting for you, @emmanuelmathot

PR #583, opened 2026-02-03
remove suggestion to use asset keys as file name

### LINKED ISSUE #579: Remove suggestion to use asset key as file name
https://github.com/Open-EO/openeo-api/blob/7483918656bc10692ce8fa9b5994dbaf2e3069be/openapi.yaml#L4596-L4600

> The keys MAY be used by clients as file names

I'd propose to remove this suggestion. 
It's a weak ("may") suggestion, so in practice,  clients can not (and consequently should not)  depend on the backend providing keys that are safe to use as file name. 

refs:
- from https://github.com/Open-EO/openeo-geopyspark-driver/issues/291#issuecomment-1352879782:
  > Adding this sentence above to the spec might not have been the wisest decision

- https://github.com/Open-EO/openeo-python-client/issues/820

soxofaan:
fixes  #579

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Remove suggestion to use (STAC) asset keys directly as file name [#579](https://github.com/Open-EO/openeo-api/issues/579)
+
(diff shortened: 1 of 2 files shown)


m-mohr (approved):
LGTM. Will merge once I have the 1.3.0 release done.

Issue #445, opened 2022-04-05
Publish maturity level of processes + collections


jdries:
As a backend, we often want to publish processes and collections to our production endpoints, even if they are very immature.
(For instance, if a beta user wants this, or simply if it's included in a software version that also carries more stable updates, or a project milestone demands it.)

On the user side, we also see users that are looking for an operational service (the default), whereas others like or have to experiment with things.

If backends advertise a maturity level, the aggregator can expose endpoints that filter on such a specific level. When the same collection is available on multiple backends, but one of them is marked as immature, it might also prefer the mature one.

There's currently an 'experimental' flag, but perhaps we need more categories? (Like 3 to 4?)

Proposal: prototype - incubating - integration - mature

m-mohr:
I can see why this seems like a good idea from the implementation/service perspective in a federated platform. On the other hand, I'm not sure users would understand the differences. Actually, I'm not really sure either. Or is this meant to be only to be used by the aggregator? Also, would back-ends use them consistently? Only then this would be useful.

The question is whether there's enough benefit to add another flag to split into 3 or 4 categories while there is already the experimental flag? Experimental could be good enough from the user perspective to show a "warning" and then you can always add a more fine-grained classification already: either through the categories/keywords (for machines) and/or through the description (for users).

jdries:
So users would indeed not see this, it would be hidden behind the aggregator. Backends wanting to join a federation would have to comply with the rules or would not be added.

Maybe we can indeed start out with the 'experimental' flag...

PR #591, opened 2026-05-21
Allow collections to be flagged as experimental

soxofaan:
Most resources (processes, file formats, service types, runtimes, ...) can be flagged as experimental, but that's not possible with collections apparently.

Not sure if this is intended for some reason (STAC compatibility?) or an accidental oversight

This PR tries to fix that

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Collections can be flagged as `experimental`
+
 ### Deprecated
 
 ### Changed
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -4398,6 +4398,8 @@ components:
             This property REQUIRES to add `version` (STAC < 1.0.0-rc.1) or
             `https://stac-extensions.github.io/version/v1.2.0/schema.json` (STAC >= 1.0.0-rc.1)
             to the list of `stac_extensions`.
+        experimental:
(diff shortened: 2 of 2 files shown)


soxofaan:
Ok, I found some pointers to discussions on this topic:

- https://github.com/Open-EO/openeo-api/issues/445
- https://github.com/openEOPlatform/architecture-docs/issues/323
- https://github.com/openEOPlatform/documentation/pull/70
- https://github.com/openEOPlatform/architecture-docs/issues/325

But most of these tickets are non-public, so hard to discover.
Most of the discussion was also in the context of openEO Platform, while the value of this flag has a larger scope than that.


Regardless, one of the public outcomes (currently at https://docs.openeo.cloud/federation/backends/collections.html#collection-availability):

> The 'experimental' flag from https://stac-extensions.github.io/version/v1.1.0/schema.json needs to be set on experimental collections. A collection can be experimental either because there are issues with the actual data or catalog, or because of backend specific issues that make the use of the collection unstable. A collection marked as experimental does not need to comply with further requirements


So in conclusion, the "experimental" flag can already be added through the "version" STAC extension, so the need to explicitly add it in the openEO API spec too is pretty low. 
Counter-point: the "deprecated" flag is also part of the "version" STAC extension, while it is also explicitly listed in the openEO API spec, so there is a bit of asymmetry here.

m-mohr:
We can add it to the spec, I think. We use it in other areas as well. The Web Editor even makes use of it already.

m-mohr (approved):

PR #593, opened 2026-06-03
Align collection "license" with STAC 

### LINKED ISSUE #592: Further align with STAC 1.1
The "license" field of collection metadata is currently:

> Alternatively, use `proprietary` if the license is not on the SPDX license list or `various` if multiple licenses apply.

The spec of STAC collections currently has (since https://github.com/radiantearth/stac-spec/commit/7c2fad3cf714d49c24ffa09b55ef4c98f0eb110e):

> String with the value `other` if the license is not on the SPDX license list. The strings `various` and `proprietary` are deprecated.

probably needs some alignment

soxofaan:
Proposal to fix #592

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Remove suggestion to use (STAC) asset keys directly as file name [#579](https://github.com/Open-EO/openeo-api/issues/579)
+- Align collection `license` field with current STAC spec: use `other` instead of deprecated `proprietary`/`various` [#592](https://github.com/Open-EO/openeo-api/issues/592)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


m-mohr:
Isn't the whole API specification still pretty much on STAC v1.0? We probably would want to go through all parts of the spec that are STAC related and make it 1.0 and 1.1 compliant.

soxofaan:
good point

> Isn't the whole API specification still pretty much on STAC v1.0?

I'm not sure it's clear that the whole API specification is on STAC v1.0. 
For example, the example value for `stac_version` is 1.1.0
https://github.com/Open-EO/openeo-api/blob/1881dae18b3c2c417f1305774cf295c81d60d400/openapi.yaml#L4584-L4592

And the generic "EO Data Discovery" section keeps it generic to 1.x.x
https://github.com/Open-EO/openeo-api/blob/1881dae18b3c2c417f1305774cf295c81d60d400/openapi.yaml#L523-L525


But indeed, `/collections` and `/collections/{collection_id}` indeed explicitly refer to STAC API 1.0.0
https://github.com/Open-EO/openeo-api/blob/1881dae18b3c2c417f1305774cf295c81d60d400/openapi.yaml#L1259

side-question: does "STAC API" here actually means what is now referred to as "STAC API", or does it mean just static "STAC"?

